PR #24, opened 2020-05-24
Sdks 250 request interceptor

jhoongo:


jhoongo:
> Overall implementation looks good to me. Good test coverage. Do you think is possible provide some sample Interceptors in the app as Android is doing?

Good point, I added an example RequestInterceptor in sample app.

PR #37, opened 2020-07-12
Sdks 494 499 attribute input callback

jhoongo:


FRAuth/FRAuth/Callbacks/NumberAttributeInputCallback.swift
```diff
@@ -0,0 +1,28 @@
...
+ NumberAttributeInputCallback is a representation of OpenAM's NumberAttributeInputCallback to collect double value with OpenAM validation and given policies.
+ */
+@objc(FRNumberAttributeInputCallback)
+public class NumberAttributeInputCallback: AttributeInputCallback {    
+    
+    /// Sets Double input value for NumberAttributeInputCallback.
+    /// - Parameter val: Double input value for Callback
+    /// - Returns: Boolean indicator whether or not it was successful
+    @objc(setNumberValue:)
+    @discardableResult public func setValue(_ val: Double) -> Bool {
```
rodrigoareis, line 24: Android implementation is not returning a boolean indicating the success of the operation. Do you think it is really necessary? Should we align the behaviour?

jhoongo: You are correct, this discardableResult of setValue methods in Callback were designed for the other approach that I was thinking, and it was not removed. There is no point with the current design to return the boolean result of the operation. Thanks for pointing that out; removed the methods, and committed.

PR #38, opened 2020-07-12
Sdks 436 device profile ui

jhoongo:


FRUI/FRUI/ViewControllers/AuthStepViewController.swift
```diff
@@ -126,7 +126,33 @@ class AuthStepViewController: UIViewController {
...
+                else {
+                    var deviceProfileCallback: DeviceProfileCallback?
+                    for callback in self.authCallbacks {
+                        if let thisCallback = callback as? DeviceProfileCallback {
+                            deviceProfileCallback = thisCallback
+                        }
+                    }
+                    //  If DeviceProfileCallback is found as one of Callbacks, collect data first
+                    if let deviceProfileCallback = deviceProfileCallback {
+                        self.startLoading()
```
cerebrl, line 145: May I ask why you're blocking the `renderAuthStep` until after you have collected the device profile?

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
```diff
@@ -126,7 +126,33 @@ class AuthStepViewController: UIViewController {
...
+                    deviceProfileCallback.execute { (profile) in
+                        self.submitNode()
+                    }
+                }
+                else {
+                    var deviceProfileCallback: DeviceProfileCallback?
+                    for callback in self.authCallbacks {
+                        if let thisCallback = callback as? DeviceProfileCallback {
+                            deviceProfileCallback = thisCallback
+                        }
```
cerebrl, line 141: This block of code is mind-bending for me O_o 

Does this mean the following:

Within this conditional block, assign `callback` to `thisCallback`. Evaluate if `callback` is of type `DeviceProfileCallback`, and if it is, assign the value of `thisCallback` (which was `callback`) to the variable `deviceProfileCallback`.

If the above is true, should I assume conditionals in Swift require a `let` statement to assign the value being evaluated to something? Sorry, if that's a dumb question :)

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
jhoongo: @cerebrl , renderAuthStep basically renders all Callback in the Node in the list view. If the DeviceProfileCallback is found as one of Callbacks in the list, then from line #144 - #150, what it does is to collect DeviceProfile information first before render all Callback, and upon completion, it renders the UI.

The reason for that is, as DeviceProfile collecting operation is async and the submit action can basically happen at anytime after the UI is rendered, I wanted to make sure that the device information is collected before submitting the node by the user.

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
jhoongo: @cerebrl , this is more of Swift syntax, and it maybe difficult to understand as we haven't seen such thing in other language. So, let me explain line by line.

So, in line#137, I declare the optional variable named, deviceProfileCallback which is currently null. 
In line#138, I'm looping through list of Callbacks in Node. 
In line#139, I'm checking if the callback in the list is DeviceProfileCallback; this may look strange, but as Node contains an array of Callback (which is super class of all Callback), I need to typecast it to DeviceProfileCallback, and that's essentially what this if statement does. You can also read it as "if this callback is DeviceProfileCallback".
In line#140, I'm assigning the DeviceProfileCallback to deviceProfileCallback variable that I declared in line#137. The reason that I keep it as variable is similar to what I answered above in the other comment. As DeviceProfile collecting operation is async, in iOS, I try to avoid running this in UI thread (and showing loading indicator when it runs in background), and ensure that the operation is completed before I render UI. So, I'm using this variable later part of the codes to collect DeviceProfile.

In short, Swift requires let statement to assign the value being evaluated to something else.  This let statement unwraps the variable and/or typcast the variable as needed. Please let me know if you have any other question. I would love to discuss about it if you want.

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
cerebrl: Thanks for the explanation, @jhoongo. You bring up a good point about the possible race condition. Are you rendering any kind of spinner or "waiting ... " message since the user may be presented with a blank screen?

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
cerebrl: Thanks for the help understanding the code, @jhoongo! I think I get it. I'll let this settle in my mind for a bit :)

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
spetrov: From my testing, no spinner is displayed... but this is expected per he acceptance criteria of SDKS-436:
"Don't display "loading indicator" / "spinner""
"Don't display message"

However, as we all agree, it is much safer to wait for the data collection to finish and then render the form...

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
```diff
@@ -126,7 +128,36 @@ class AuthStepViewController: UIViewController {
                 //  Set auth callbacks in list for rendering
                 self.currentNode = node
                 self.authCallbacks = node.callbacks
-                self.renderAuthStep()
+                
+                //  If DeviceProfileCallback is the only Callback, then immediately collect, and submit
+                if self.authCallbacks.count == 1, let deviceProfileCallback = self.authCallbacks.first as? DeviceProfileCallback {
+                    self.loadingView.loadingText = deviceProfileCallback.message
```
spetrov, line 134: I was never able to see the configured message in AM showing up with the spinner, although I see that the loadingText is set with correct value, etc... What I see is the spinner with with message "Loading..."

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
jhoongo: @spetrov , I've updated the code to better visualize loading indicator when there is only DeviceProfileCallback. Instead of having a generic loading indicator that goes away quickly, I re-render the TableView with dedicated loading view cell for DeviceProfileCallback. Please let me know if this makes better sense. Thanks!

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
jhoongo: @spetrov I've updated the loading indicator to better visualize the situation. Please pull the latest code, and let me know if that works better for you :)

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
spetrov: Yes, it's much better now! Thanks!

PR #44, opened 2020-08-05
Sdks 504 email suspended node

jhoongo:


FRAuth/FRAuth/Authentication/AuthService.swift
```diff
@@ -18,7 +18,7 @@ import FRCore
     * An error, if occurred during the authentication flow
  
  ## Notes ##
-     * Any Callback type returned from OpenAM must be supported within CallbackFactory.shared.supportedCallbacks.
+     * Any Callback type returned from AM must be supported within CallbackFactory.shared.supportedCallbacks.
```
rodrigoareis, line 21: Update year on the header for modified files?

jhoongo: Thanks for pointing that out @rodrigoareis . Updated all files in this PR.

FRAuth/FRAuth/FRAuth.swift
```diff
@@ -228,6 +228,18 @@ public final class FRAuth: NSObject {
     
     
     // - MARK: Private
+    
+    /// Initiates Authentication Tree with `suspendedId`
+    /// - Parameters:
+    ///   - suspendedId: suspendedId contained in resumeURI contained in Email received from `Email Suspend Node` in AM
+    ///   - completion: NodeCompletion callback which returns the result of Node submission
+    func next<T>(suspendedId: String, completion:@escaping NodeCompletion<T>) {
```
cerebrl, line 236: Should there be a space between the `completion:` and `@escaping` ... I'm guessing not as I see no space multiple times throughout the codebase, but I thought I'd mention it.

FRAuth/FRAuth/Session/FRSession.swift
```diff
@@ -99,6 +99,30 @@ import Foundation
     }
     
     
+    /// Invokes /authenticate endpoint in AM to go through Authentication Tree flow with  `resumeURI` and `suspendedId` to resume Authentication Tree flow.
+    /// - Parameters:
+    ///   - resumeURI: Resume URI received in Email from Suspned Email Node; URI **must** contain `suspendedId` in URL query parameter
```
cerebrl, line 104: typo: `Suspned`

FRAuth/FRAuth/Authentication/AuthService.swift
```diff
@@ -2,7 +2,7 @@
 //  AuthService.swift
 //  FRAuth
 //
-//  Copyright (c) 2019 ForgeRock. All rights reserved.
+//  Copyright (c) 2020 ForgeRock. All rights reserved.
```
cerebrl, line 5: I think we are supposed to do a range: `<initial publish year> - <current year>`

SampleApps/FRExample/FRExample/AppDelegate.swift
```diff
@@ -23,6 +24,36 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+                var currentController = rootViewController
+                while let presentedController = currentController.presentedViewController {
+                    currentController = presentedController
+                }
+                FRSession.authenticateWithUI(resumeURL, currentController) { (token: Token?, error) in
+                    if let error = error {
+                        FRLog.e(error.localizedDescription)
+                    }
+                    else {
+                        FRLog.i("Authenticate with ResumeURI successful: \(String(describing: token))")
```
cerebrl, line 49: This may be a silly question, but is the `\` intended here?

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
```diff
@@ -348,6 +356,20 @@ class AuthStepViewController: UIViewController {
         self.view.endEditing(true)
         self.dismiss(animated: true, completion: nil)
     }
+    
+    @IBAction func closeButtonClicked(sender: UIButton) {
+        if let completion = self.atCompletion {
```
cerebrl, line 361: Why the 3 separate conditions if they all result in the same function call?

FRAuth/FRAuth/Authentication/AuthService.swift
```diff
@@ -73,6 +78,25 @@ public class AuthService: NSObject {
...
+    /// - Parameters:
+    ///   - suspendedId: suspendedId to resume Authentication Tree flow
+    ///   - serverConfig: ServerConfig object for AuthService server communication
+    ///   - oAuth2Config: OAuth2Client object for AuthService OAuth2 protocol upon completion of authentication flow, and when SSOToken received, AuthService automatically exchanges the token to OAuth2 token set
+    ///   - sessionManager: SessionManager instance to manage and persist authenticated session
+    ///   - tokenManager: TokenManager  instance to manage and persist authenticated session
+    init(suspendedId: String, serverConfig: ServerConfig, oAuth2Config: OAuth2Client?, sessionManager: SessionManager? = nil, tokenManager: TokenManager? = nil) {
+        FRLog.v("AuthService init - suspendedId: \(suspendedId), ServerConfig: \(serverConfig), OAuth2Client: \(String(describing: oAuth2Config)), SessionManager: \(String(describing: sessionManager)), TokenManager: \(String(describing: tokenManager))")
+        self.serviceName = suspendedId
+        self.authIndexType = OpenAM.suspendedId
```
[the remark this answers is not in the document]
jhoongo: Yes, that's correct, OpenAM.suspendedId is just a constant property that I manage in OpenAM. The reason for assigning OpenAM.suspendedId to authIndexType is due to the structure of AuthService. Currently, AuthService has authIndexType, and serviceName (which needs to be renamed to authIndexValue). For Email Suspend Node flow, we only need authIndexValue, and need to skip the authIndexType. With the current structure of AuthService, it wasn't ideal to have additional property of suspendedId; it's more of Swift syntax, but in short, properties in Swift class can be either optional, or mandatory, and if it is defined as mandatory, I need to set the value. So, if I add suspendedId property to AuthService class, then serviceName and authIndexType will eventually have to have some value. If I put these properties as an optional values, then I need to more validations. That's why I'm re-using the serviceName, and authIndexType properties as it is, and setting the authIndexType with constant string value to differentiate the flow. Hope this makes sense, and let me know if you have any further question.

FRAuth/FRAuth/Error/AuthApiError.swift
```diff
@@ -0,0 +1,121 @@
+//
```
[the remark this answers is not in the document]
jhoongo: Yes, that's correct. It's to align with what we have discussed on Fri, and to align with Android SDK's exception handling to separate AM API exceptions from other exceptions. I will document more on the iOS SDK's error structure and types in the Wiki page.

FRAuth/FRAuth/FRAuth.swift
jhoongo: That's right. Although we do not currently use any linting tool, that's one of my objects in the future to get rid of this kind of issues. Thanks for pointing that out!

FRAuth/FRAuth/Session/FRSession.swift
jhoongo: Thanks, will update this.

FRAuth/FRAuth/Authentication/AuthService.swift
jhoongo: I just saw the thread that that would be ideal. So will update all files accordingly. Thanks for pointing that out.

SampleApps/FRExample/FRExample/AppDelegate.swift
jhoongo: It's the way of Swift interpolation. Basically whenever you would like to print out a variable within String other than reconstructing the string, you can use \() to inject variable.

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
jhoongo: This was designed way back when I started the project and had very little knowledge of handling the generic types in Swift. This is something in my to-do list for refactoring a little bit in UI codebase. Essentially the reason for this conditions were to handle three different callback types (returning Token, AccessToken, and FRUser objects) in a single unified page.

FRAuth/FRAuth/Authentication/AuthService.swift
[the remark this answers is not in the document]
cerebrl: This makes sense. Thanks for the explanation!

SampleApps/FRExample/FRExample/AppDelegate.swift
cerebrl: Interesting, never seen this type of syntax before.

cerebrl:
Looks good to me!

PR #54, opened 2020-09-21
[SDKS-673] Updating Secure Enclave validation with CryptoKit, and min…

jhoongo:
…or changes for version updates


spetrov (approved):
Looks good to me. I did test the changes locally and confirmed that secure enclave presence detection works as expected and key pairs are generated successfully on various devices (tested with iOS 10, 11, 12, 13 and 14 including iPod 7th generation)

PR #59, opened 2020-10-28
[SDKS-648] Old refresh_token is not persisted when refresh_token gran…

jhoongo:
…t does not return new refresh_token


spetrov (approved):
Looks good to me! 👍

PR #67, opened 2020-11-23
[SDKS-781] [iOS] Logout should only call either /endSession or /sessions

jhoongo:


spetrov (approved):
Looks good to me!

rodrigoareis (approved):
Changes looks good to me

PR #73, opened 2021-01-26
[SDKS-836] Expose Collector and JailbreakDetector initialization meth…

jhoongo:
…ods as public

### DIFF
--- a/FRAuth/FRAuth/Device/HardwareCollector.swift
+++ b/FRAuth/FRAuth/Device/HardwareCollector.swift
@@ -18,6 +18,9 @@ public class HardwareCollector: DeviceCollector {
     /// Name of current collector
     public var name: String = "hardware"
     
+    /// Initializes BrowserCollector instance
+    public init() { }
+    
     /// Collects hardware information using ProcessInfo
     ///
     /// - Parameter completion: completion block
--- a/FRAuth/FRAuth/Device/HardwareCollector.swift
+++ b/FRAuth/FRAuth/Device/HardwareCollector.swift
@@ -18,7 +18,7 @@ public class HardwareCollector: DeviceCollector {
     /// Name of current collector
     public var name: String = "hardware"
     
-    /// Initializes BrowserCollector instance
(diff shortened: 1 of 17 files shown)


FRAuth/FRAuth/Device/HardwareCollector.swift
```diff
@@ -18,6 +18,9 @@ public class HardwareCollector: DeviceCollector {
     /// Name of current collector
     public var name: String = "hardware"
     
+    /// Initializes BrowserCollector instance
```
cerebrl, line 21: I think you meant `HardwareCollector`, rather than `BrowserCollector` :)

cerebrl (commented):
Other than the one likely copy-paste issue, this looks good.

FRAuth/FRAuth/Device/ProfileCollector.swift
```diff
@@ -19,7 +19,7 @@ public class ProfileCollector: DeviceCollector {
     public var collectors: [DeviceCollector]
     
     /// Private initialization method which initializes default array of ProfileCollector
```
spetrov, line 21: Minor: "Private" should become "public" in comment

spetrov (approved):
Looks good to me! I only have a minor comment regarding a comment in the code that we may want to update...

FRAuth/FRAuth/Device/ProfileCollector.swift
jhoongo: Thanks Stoyan! Updated the comment accordingly.

FRAuth/FRAuth/Device/HardwareCollector.swift
jhoongo: Thanks Justin, I've updated the inline comment. :)

spetrov (approved):
👍

cerebrl (approved):

george-bafaloukas-forgerock (approved):
Looks good

PR #74, opened 2021-02-03
SDKS-806 : Removed hard-coded cookie name value from sample app / exp…

jhoongo:
…ose all configurations with default values in sample app

### DIFF
--- a/SampleApps/FRExample/FRExample/ViewController.swift
+++ b/SampleApps/FRExample/FRExample/ViewController.swift
@@ -637,8 +637,21 @@ class ViewController: UIViewController {
             return
         }
         
+        //  Default Cookie Name for SSO Token in AM
+        var cookieName = "iPlanetDirectoryPro"
+        
+        //  If custom cookie name is defined in configuration file, update the cookie name
+        if let path = Bundle.main.path(forResource: FRAuth.configPlistFileName, ofType: "plist"), let config = NSDictionary(contentsOfFile: path) as? [String: Any], let configCookieName = config["forgerock_cookie_name"] as? String {
+            cookieName = configCookieName
+        }
+        
         var request = URLRequest(url: url)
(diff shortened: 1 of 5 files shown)


george-bafaloukas-forgerock (approved):

SampleApps/FRExample/FRExample/ViewController.swift
```diff
@@ -637,8 +637,21 @@ class ViewController: UIViewController {
             return
         }
         
+        //  Default Cookie Name for SSO Token in AM
+        var cookieName = "iPlanetDirectoryPro"
+        
+        //  If custom cookie name is defined in configuration file, update the cookie name
+        if let path = Bundle.main.path(forResource: FRAuth.configPlistFileName, ofType: "plist"), let config = NSDictionary(contentsOfFile: path) as? [String: Any], let configCookieName = config["forgerock_cookie_name"] as? String {
```
cerebrl, line 644: This is kind of a nitpick, but should we have a bit higher specificity on the naming regarding the cookie? As our platforms scales in complexity, we may end up with more than one cookie to manage on the client. Using `forgerock-cookie-name` or `cookieName` _may_ force us to break backwards compatibility in the future. May I suggest `session-cookie-name` and `sessionCookieName` for better future proofing?

cerebrl (commented):
PR looks good. I just had one non-blocking comment about naming.

spetrov (approved):
Looks good to me! 👍

SampleApps/FRExample/FRExample/ViewController.swift
jhoongo: @cerebrl Thanks for the suggestion. That makes sense, and I agree with that; however, since this configuration name was introduced while ago, and is also used in Android SDK with the same name, let's discuss internally on when we can change the name.

cerebrl: Ah, good to know. Let's move this to our architecture discussion then. Thanks, @jhoongo!

PR #75, opened 2021-02-11
Sdks 873 sso token o auth2 config

jhoongo:


FRAuth/FRAuth/Authentication/Node.swift
```diff
@@ -42,26 +42,27 @@ public class Node: NSObject {
...
     ///   - authServiceId: Unique UUID for current AuthService flow
     ///   - authServiceResponse: JSON response object of AuthService in OpenAM
     ///   - serverConfig: ServerConfig object for AuthService/Node communication
     ///   - serviceName: Service name for AuthService (TreeName)
     ///   - authIndexType: authIndexType value in AM (default to 'service')
     ///   - oAuth2Config: (Optional) OAuth2Client object for AuthService/Node communication for abstraction layer
-    /// - Throws: AuthError error may be thrown from parsing AuthService response, and parsing Callback(s)
-    init?(_ authServiceId: String, _ authServiceResponse: [String: Any], _ serverConfig: ServerConfig, _ serviceName: String, _ authIndexType: String, _ oAuth2Config: OAuth2Client? = nil, _ sessionManager: SessionManager? = nil, _ tokenManager: TokenManager? = nil) throws {
+    ///   - keychainManager: KeychainManager instance to manage and persist authenticated session
+    ///   - tokenManager: TokenManager  instance to manage and persist authenticated session
```
cerebrl, line 63: These comments make it seem like the `keychainManager` and `tokenManager` do the exact same thing. Is that intentional?

FRAuth/FRAuth/Authentication/Node.swift
```diff
@@ -203,7 +211,7 @@ public class Node: NSObject {
                             }
                             else {
                                 if let token = accessToken {
-                                    try? self.sessionManager?.setAccessToken(token: token)
+                                    try? self.keychainManager?.setAccessToken(token: token)
```
george-bafaloukas-forgerock, line 214: Super minor, but you could make it "let _ = try? self.keychainManager?.setAccessToken(token: token)" to get read of the warning

FRAuth/FRAuth/Manager/TokenManager.swift
```diff
@@ -209,9 +192,9 @@ struct TokenManager {
     /// - Parameter completion: Completion block which will return an Error if there was any error encountered
     func revoke(completion: @escaping CompletionCallback) {
         do {
-            if let token = try self.retrieveAccessTokenFromKeychain() {
+            if let token = try self.keychainManager.getAccessToken() {
                 self.oAuth2Client.revoke(accessToken: token, completion: completion)
-                try? self.sessionManager.setAccessToken(token: nil)
+                try? self.keychainManager.setAccessToken(token: nil)
```
george-bafaloukas-forgerock, line 197: Since this is already in a do catch block is there a reason for that to not throw and use the try?

george-bafaloukas-forgerock (dismissed):
Looks good, a couple really minor findings mostly for discussion

FRAuth/FRAuth/Authentication/Node.swift
jhoongo: Thanks for pointing that out! I missed that part, and wasn't intentional. Just updated the description.

FRAuth/FRAuth/Authentication/AuthService.swift
```diff
@@ -69,7 +69,7 @@ public class AuthService: NSObject {
     ///   - name: String value of AuthService name
     ///   - serverConfig: ServerConfig object for AuthService server communication
     @objc
-    public init(name: String, serverConfig:ServerConfig) {
+    public init(name: String, serverConfig: ServerConfig) {
         FRLog.v("AuthService init - service: \(name), ServerConfig: \(serverConfig)")
         self.serviceName = name
         self.authIndexType = "service"
```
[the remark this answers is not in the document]
jhoongo: Nice catch, thanks for pointing that out.

FRAuth/FRAuth/Authentication/Node.swift
jhoongo: Yes, I've updated all compiler warnings to correctly handle the exception from the try/catch block. We should try to catch most of exceptions to provide more information to the developers. It wasn't really minor, but was really important catch. Thanks.

FRAuth/FRAuth/Manager/TokenManager.swift
jhoongo: Updated as per previous comment.

george-bafaloukas-forgerock (approved):

PR #76, opened 2021-02-25
Sdks 795 x code build fix

jhoongo:


spetrov (approved):
Looks good to me. 👍 
We confirmed that the project builds on older versions where the CryptoKit library is not available.

PR #79, opened 2021-03-11
Added new build scheme for BitBar integration needs

spetrov:
This pull requests adds a new build scheme into the iOS-SDK project called "FRTestHostBitBar". 
The main difference of this scheme from the original "FRTestHost" scheme is that this one includes only the FRAuthTests. This significantly decreases the size of the test app (.ipa) which we upload to BitBar.


george-bafaloukas-forgerock (approved):
Looks good

jhoongo (approved):

PR #80, opened 2021-03-19
Sdks 879 social login

jhoongo:


FRAuth/FRAuth/Callbacks/IdPCallback.swift
```diff
@@ -0,0 +1,207 @@
...
+//
+
+import Foundation
+import UIKit
+
+
+/**
+ IdPCallback is a representation of `Social Provider Handler Node` in AM when `Client Type` option specified as `NATIVE` (only available in AM 7.1 and above)
+ */
+public class IdPCallback: MultipleValuesCallback {
```
george-bafaloukas-forgerock, line 18: Can we make this conform to Codable in order to make the JSON serialisation a bit cleaner/easier

FRAuth/FRAuth/Callbacks/IdPCallback.swift
```diff
@@ -0,0 +1,207 @@
...
+        var idpHandler: IdPHandler?
+        if provider.lowercased().contains("apple") {
+            idpHandler = AppleSignInHandler()
+        }
+        else if provider.lowercased().contains("google") {
+            if let c: NSObject.Type = NSClassFromString("FRGoogleSignIn.GoogleSignInHandler") as? NSObject.Type, let thisHandler = c.init() as? IdPHandler {
+                idpHandler = thisHandler
+            }
+        }
+        else if provider.lowercased().contains("facebook") {
```
george-bafaloukas-forgerock, line 184: I would suggest creating a providers enum to get the raw values to compare, instead of using a static string value

FRAuth/FRAuth/Callbacks/SelectIdPCallback.swift
```diff
@@ -0,0 +1,65 @@
...
+    /// An array of Identity Provider value
+    public var providers: [IdPValue]
+    
+    
+    //  MARK: - Init
+    
+    /// Designated initialization method for SelectIdPCallback
+    /// - Parameter json: JSON object of SelectIdPCallback
+    /// - Throws: AuthError.invalidCallbackResponse for invalid callback response
+    public required init(json: [String : Any]) throws {
```
george-bafaloukas-forgerock, line 31: Same question regarding Codable

FRFacebookSignIn/FRFacebookSignIn/Sources/FacebookSignInHandler.swift
```diff
@@ -0,0 +1,89 @@
...
+        }
+    }
+    
+    
+    /// Generates, and returns `UIView` for `FBLoginButton` button
+    /// - Returns: `FBLoginButton` button in `UIView`
+    public func getProviderButtonView() -> UIView? {
+        let btn = FBLoginButton()
+        btn.loginTracking = .limited
+        btn.tooltipColorStyle = .neutralGray
```
george-bafaloukas-forgerock, line 70: Just a suggestion for discussion, it might be worth having the loginTracking and tooltipColorStyle as input params in this function to allow client customization

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
```diff
@@ -255,6 +263,22 @@ class AuthStepViewController: UIViewController {
             return
         }
         
+        var listItems: [Any] = []
+        for callback in self.authCallbacks {
+            if let idpCallback = callback as? SelectIdPCallback {
+                for provider in idpCallback.providers {
+                    if provider.provider != "localAuthentication" {
```
george-bafaloukas-forgerock, line 270: I would suggest against comparing against hardcoded strings. If there is no better way to identify the provider then maybe we should have that as part of a Providers enum helper

FRUI/FRUI/Views/IdPValueTableViewCell.swift
```diff
@@ -0,0 +1,83 @@
...
+    
+    //  MARK: - Protocols
+    
+    func updateCellData(callback: Callback) {
+        
+    }
+    
+    func updateCellWithProvider(provider: IdPValue) {
+        
+        if provider.provider.lowercased().contains("apple") {
```
george-bafaloukas-forgerock, line 44: See pervious comment regarding string values

FRAuth/FRAuth/Callbacks/IdPCallback.swift
rodrigoareis: That's a good idea. I believe we could make the ```Callback``` class ```Codable```, which would be inherited by all callback classes

FRAuth/FRAuth/Constants/Typealias.swift
```diff
@@ -33,6 +33,12 @@ public typealias CompletionCallback = (_ error:Error?) -> Void
 public typealias JSONCompletionCallback = (_ result: [String: Any]) -> Void
 
 
+//  MARK: - Social Login
```
rodrigoareis, line 36: Update Copyright header

SampleApps/FRExample/FRExample/AppDelegate.swift
```diff
@@ -11,6 +11,12 @@
 import UIKit
 import FRAuth
 import FRCore
+#if canImport(FRFacebookSignIn)
```
rodrigoareis, line 14: Update Copyright header

SampleApps/FRExample/FRExample/AppDelegate.swift
```diff
@@ -27,6 +33,17 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     
     func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey : Any] = [:]) -> Bool {
         
+        #if canImport(FRFacebookSignIn)
```
george-bafaloukas-forgerock, line 36: Is this the wrong way around? Shouldn't it be:

#if canImport(FRFacebookSignIn)
        if FacebookSignInHandler.handle(app, url, options) {
            return true
        }
        #endif
        #if canImport(FRGoogleSignIn)
        
        if GoogleSignInHandler.handle(app, url, options) {
            return true
        }
        #endif

george-bafaloukas-forgerock: Sorry about the formatting but it really doesn't like me pasting the code :)

FRAuth/FRAuth/Callbacks/IdPCallback.swift
jhoongo: @george-bafaloukas-forgerock , @rodrigoareis , I do agree that it would be much cleaner if we were to do it with Codable; however, if I recall correctly, I had tried with Codable approach when I initially implement Callback structure, and encountered some issue with Objc compatibility or something else. It is still worthwhile to investigate further on that topic, but as this would take entire Callback structure changes, are you guys okay with having a separate story for the investigation?

Thanks,

FRAuth/FRAuth/Callbacks/IdPCallback.swift
jhoongo: @george-bafaloukas-forgerock Creating an enumeration value for providers may not make sense in this case as this provider value is plain text input from the server side. One thing that I have not implemented in here is the Callback constants that I had implemented it for other feature you've tested and since that feature branch is still in stash, it is not merged and applied to this feature implementation. 

I would suggest instead of having enum value for the provider value, I prefer to have Callback constants value to compare those three providers.

Let me know what you think.

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
jhoongo: Same thing here. IdPValue.provider is a plain text input from the end-user, so enum may not be helpful. For this implementation, I have not incorporated the Callback constants that you've suggested for the previous feature testing. I am planning to apply that constant values later when I merge this repo into the stash repo.

SampleApps/FRExample/FRExample/AppDelegate.swift
jhoongo: @george-bafaloukas-forgerock Thanks for pointing this out.

jhoongo: Updated

SampleApps/FRExample/FRExample/AppDelegate.swift
jhoongo: Updated

FRAuth/FRAuth/Constants/Typealias.swift
jhoongo: Updated

FRFacebookSignIn/FRFacebookSignIn/Sources/FacebookSignInHandler.swift
jhoongo: Updated

FRAuth/FRAuth/Callbacks/IdPCallback.swift
george-bafaloukas-forgerock: Hey James, I do agree that this would require a significant amount of time to refactor. A spike to investigate how much this time would be, if we have any issue and what other classes would benefit from such refactoring is a good idea in my opinion

FRAuth/FRAuth/Callbacks/IdPCallback.swift
george-bafaloukas-forgerock: Callback constants is OK with me.

FRUI/FRUI/ViewControllers/AuthStepViewController.swift
george-bafaloukas-forgerock: Same as above Callback constants is OK with me.

rodrigoareis (approved):
changes looks good to me

george-bafaloukas-forgerock:
@jhoongo I am happy with changes. Have we created a ticket for the 2 discussions that will have to be addressed separately? 
1. The static Strings 
2. Investigating the use of Codable

spetrov (dismissed):
Looks good to me! 👍

rodrigoareis (approved):
Changes looks good to me

FRUI/FRUI/Views/IdPCallbackTableViewCell.swift
```diff
@@ -0,0 +1,59 @@
...
+
+class IdPCallbackTableViewCell: UITableViewCell, FRUICallbackTableViewCell {
+
+    // MARK: - Properties
+    
+    public static let cellIdentifier = "IdPCallbackTableViewCellId"
+    public static let cellHeight: CGFloat = 54.0
+    public var presentingViewController: UIViewController?
+    var callback: IdPCallback?
+    var delegate: AuthStepProtocol?
```
george-bafaloukas-forgerock, line 22: Shouldn't that be weak?

FRUI/FRUI/Views/IdPValueTableViewCell.swift
```diff
@@ -0,0 +1,83 @@
...
+
+import UIKit
+import FRAuth
+
+class IdPValueTableViewCell: UITableViewCell, FRUICallbackTableViewCell {
+
+    //  MARK: - Properties
+    public static let cellIdentifier = "IdPValueTableViewCellId"
+    public static let cellHeight: CGFloat = 54.0
+    var delegate: AuthStepProtocol?
```
george-bafaloukas-forgerock, line 20: Same as above, should the delegate be weak?

spetrov (approved):
Approved!

PR #83, opened 2021-04-21
SDKS-993 Image attribute in QRCode not persisted by SDK

rodrigoareis:
This fix address an issue on persist the `image` attribute from a QRCode for OATH registration. The SDK was expecting the attribute to be Base64 encoded. It's only the case for Push accounts, OATH accounts usually has the image value as URL encoded or plain text. Support for image value as Base64 was kept as well.

Unit tests were added to test the changes.


jhoongo (approved):
looks good to me

george-bafaloukas-forgerock (approved):
looks good

spetrov (approved):
Looks good to me!

PR #88, opened 2021-05-04
SDKS-910-Remove_deprecated_methods

george-bafaloukas-forgerock:
- Removed ServerConfig deprecated public initializer
- Deleting deprecated FRAuth.shared.next() (public func next<T>) method from FRAuth.swift
- Removed FRURLProtocol.validatedURLs from FRURLProtocol.swift
- Removed deprecated FRURLProtocolResponseEvaluationCallback and dependencies from Typealias.swift, FRURLProtocol.swift, TokenManagementPolicy.swift
- Removed public var value from SingleValueCallback. The var has been replaced by getValue and setValue methods

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,10 +15,14 @@
 - `PlatformCollector`'s attribute names were changed from `timezone` and `jailbreakScore` to `timeZone` and `jailBreakScore` respectively to align with AM and Android SDK. [SDKS-908]
 - `Browser.validateBrowserLogin()` is now available in Objective-c as well. [SDKS-975]
 - Jailbreak detection logic was updated to prevent Jailbreak detection bypass. [SDKS-840]
-- Removed previously deprecated methods. [SDKS-910]
+
 
 #### Deprecated
--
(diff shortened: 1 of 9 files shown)


jhoongo (changes requested):
In general, looks good. Few comments:

1.  In DeviceCollectorNodeTests.swift line#813, it looks like this particular test is still using the .value property of Callback. This was supposed to be removed, but I might have overlooked this when I was refactoring all the test codes. It would be nice if you could handle this as this would throw an exception for building test.
2. CHANGELOG.md, the deprecation notice would be better to go under Deprecated section, and it would be nice if we can just list interfaces that are completely removed.

spetrov (commented):
I agree with James that the removed methods should be listed in the "Deprecated" section below. Other than that everything looks good to me!

george-bafaloukas-forgerock:
Hey @spetrov & @jhoongo, please have another look, comments addressed. Apologies while I did run the FRAuth tests I did miss those cases.

jhoongo (approved):

jhoongo:
Looks good to me

PR #101, opened 2021-08-06
SDKS-1157-FRExample - SDK crashes when attempting Centralised Login

george-bafaloukas-forgerock:
iOS 14.5 has introduced a change not allowing the callBackURLScheme to contain the full URI but requires only the scheme (ex: frauth not frauth://).

Passing the redirectUri.scheme instead of full URL solved the issue


spetrov (approved):
LGTM 👍🏻

cerebrl (approved):

PR #104, opened 2021-09-01
SDKS-1209-Metadata callback overrides stage property of node

george-bafaloukas-forgerock:
Checking if the stage property has been set before going through the MetadataCallback to set the stage of the node.


rodrigoareis (dismissed):
Change looks good to me

spetrov (approved):
LGTM 👍🏻

spetrov (approved):

PR #106, opened 2021-09-03
SDKS-1241 Align identity validation in QRCode parsing for OATH

rodrigoareis:


FRAuthenticator/FRAuthenticator/Model/Mechanism/OATH/OathQRCodeParser.swift
```diff
@@ -155,5 +155,11 @@ struct OathQRCodeParser {
         if self.issuer.count == 0 {
             self.issuer = self.label
         }
+        
+        if self.issuer == "" && self.label == "" {
```
george-bafaloukas-forgerock, line 159: Instead of assigning empty values I would suggest making those properties optional if they can be empty. Unless, there is a business reason not to.

george-bafaloukas-forgerock: Furthermore, in various places we have comparisons against static strings:
Ex: item.name == "algorithm"

I would suggest for safety and future proofing to move those static strings as part of an Enum or a "Values" static helper file

george-bafaloukas-forgerock (approved):
Approved with suggestions

spetrov (approved):
LGTM 👍🏻

FRAuthenticator/FRAuthenticator/Model/Mechanism/OATH/OathQRCodeParser.swift
rodrigoareis: @george-bafaloukas-forgerock I agree with all suggested changes. However, I would suggest handle those in a future release.

PR #110, opened 2021-11-02
SDKS-1334_Change-SecuredKey-accessibility

george-bafaloukas-forgerock:
Updated SecuredKey to get the accessibility flag as a property and match the one that the keychainService uses

Updated tests & copyright

### DIFF
--- a/FRCore/FRCore/Keychain/SecuredKey.swift
+++ b/FRCore/FRCore/Keychain/SecuredKey.swift
@@ -2,7 +2,7 @@
 //  SecuredKey.swift
 //  FRCore
 //
-//  Copyright (c) 2020 ForgeRock. All rights reserved.
+//  Copyright (c) 2020 - 2021 ForgeRock. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
@@ -23,6 +23,8 @@ public struct SecuredKey {
     fileprivate var publicKey: SecKey
     /// Algorithm to be used for encryption/decryption using SecuredKey
     fileprivate let algorithm: SecKeyAlgorithm = .eciesEncryptionCofactorX963SHA256AESGCM
+    
+    public var kSecAttrAccessibleFlag: KeychainAccessibility = .afterFirstUnlock
     
     /// Validates whether SecuredKey using Secure Enclave is available on the device or not
     public static func isAvailable() -> Bool {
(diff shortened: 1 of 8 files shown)


FRCore/FRCore/Keychain/SecuredKey.swift
```diff
@@ -60,7 +60,7 @@ public struct SecuredKey {
     
     /// Initializes SecuredKey object with designated service; SecuredKey may return nil if it failed to generate keypair
     /// - Parameter applicationTag: Unique identifier for SecuredKey in Keychain Service
-    public init?(applicationTag: String, accessGroup: String? = nil) {
+    public init?(applicationTag: String, accessGroup: String? = nil, kSecAttrAccessibleFlag: KeychainAccessibility) {
```
rodrigoareis, line 63: I would suggest make the parameter anme match the one in the ```KeychainOptions```. Meaning rename it from ```kSecAttrAccessibleFlag``` to ```accessibility```.

Also, I am wondering if instead of forcing the developer to pass the ```accessibility``` parameter, we set the default value to ```.afterFirstUnlock```. Aligning with the behaviour we see on the ```KeychainOptions```.

rodrigoareis (commented):
Changes looks good to me. I provided some comments to be reviewed.

FRCore/FRCore/Keychain/SecuredKey.swift
spetrov: I agree with Rodrigo. We should preserve the old interface, otherwise, this becomes a "breaking" change...

spetrov: btw, because of this interface change, some of the test cases in FRCore do not compile and make the Jenkins build fail... 
The tests in question are in here: /FRCore/FRCoreTests/FRCore/Util/SecuredKeyTests.swift

They all pass once I pass "kSecAttrAccessibleFlag: .afterFirstUnlock" as parameter...

FRCore/FRCoreTests/FRCore/Util/KeychainServiceTests.swift
```diff
@@ -15,7 +15,7 @@ class KeychainServiceTests: FRBaseTestCase {
     var kc: KeychainService?
     
     override func setUp() {
-        self.configFileName = "Config-live-01"
```
spetrov, line 18: Is this change is needed?

spetrov (changes requested):

FRCore/FRCoreTests/FRCore/Util/KeychainServiceTests.swift
george-bafaloukas-forgerock: The 'Config-live-01' file is empty on mine, so the test was failing. Not sure actually if I needed to change that. I will revert

spetrov: 👍🏻 Yes, this file is supposed to be empty. It is needed for the "live" e2e test cases, which the Jenkins pipeline configure as needed...

spetrov (dismissed):
👍🏻 Everything looks good to me! 
I'd only ask to update the changelog.md (add a note about this change...)

spetrov (approved):
Looks good! 👍🏻

rodrigoareis (approved):
Looks good to me

PR #111, opened 2021-11-11
Sdks-1366_iOS make sure that deprecated "unarchiveObject(with:" methods are not used for iOS 11+

george-bafaloukas-forgerock:
OS make sure that deprecated "unarchiveObject(with:" methods are not used for iOS 11+
Based on the report from: https://forgerock.zendesk.com/agent/tickets/67734

- Added FRHTTPCookie class to support NSSecureCoding
- Fixed FRRestClient to only unarchive using NSSecureCoding methods for iOS 11+
- Fixed KeychainServiceClient to only unarchive using NSSecureCoding methods for iOS 11+

### DIFF
--- a/FRAuth/FRAuth/Network/FRHTTPCookie.swift
+++ b/FRAuth/FRAuth/Network/FRHTTPCookie.swift
@@ -31,7 +31,7 @@ class FRHTTPCookie: HTTPCookie, NSSecureCoding {
         if #available(iOS 14.0, *) {
             portList = coder.decodeArrayOfObjects(ofClass: NSNumber.self, forKey: "portList")
         } else {
-            portList = coder.decodeObject(of: [NSNumber.self], forKey: "portList") as? [NSNumber]
+            portList = coder.decodeObject(of: [NSArray.self, NSNumber.self], forKey: "portList") as? [NSNumber]
         }
         
         let sameSitePolicy = coder.decodeObject(of: NSString.self, forKey: "sameSitePolicy") as String?
@@ -41,14 +41,14 @@ class FRHTTPCookie: HTTPCookie, NSSecureCoding {
         properties[HTTPCookiePropertyKey.value] = value
         properties[HTTPCookiePropertyKey.domain] = domain
         properties[HTTPCookiePropertyKey.path] = path
-        properties[HTTPCookiePropertyKey.secure] = isSecure
+        properties[HTTPCookiePropertyKey.secure] = isSecure ? "TRUE" : nil
         properties[HTTPCookiePropertyKey.expires] = expiresDate
         properties[HTTPCookiePropertyKey.comment] = comment
         properties[HTTPCookiePropertyKey.commentURL] = commentURL
-        properties[HTTPCookiePropertyKey.discard] = isSessionOnly
+        properties[HTTPCookiePropertyKey.discard] = isSessionOnly ? "TRUE" : nil
         properties[HTTPCookiePropertyKey.maximumAge] = expiresDate
         properties[HTTPCookiePropertyKey.port] = portList
-        properties[HTTPCookiePropertyKey("HttpOnly")] = isHTTPOnly
+        properties[HTTPCookiePropertyKey("HttpOnly")] = isHTTPOnly ? "TRUE" : nil
         if #available(iOS 13.0, *) {
             if let sameSitePolicyValue = sameSitePolicy {
                 properties[HTTPCookiePropertyKey.sameSitePolicy] = HTTPCookieStringPolicy(rawValue: sameSitePolicyValue)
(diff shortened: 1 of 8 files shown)


FRAuth/FRAuth/Network/FRHTTPCookie.swift
```diff
@@ -0,0 +1,88 @@
...
+        let isSessionOnly = coder.decodeBool(forKey: "isSessionOnly")
+        let domain = coder.decodeObject(of: NSString.self, forKey: "domain") as String?
+        let path = coder.decodeObject(of: NSString.self, forKey: "path") as String?
+        let isSecure = coder.decodeBool(forKey: "isSecure")
+        let isHTTPOnly = coder.decodeBool(forKey: "isHTTPOnly")
+        
+        let comment = coder.decodeObject(of: NSString.self, forKey: "comment") as String?
+        let commentURL = coder.decodeObject(of: NSURL.self, forKey: "commentURL") as URL?
+        let portList: [NSNumber]?
+        if #available(iOS 14.0, *) {
```
lsanches, line 30: Does this mean that our SDK is not backward compatible with iOS 10.0 anymore?

FRAuth/FRAuth/Network/FRHTTPCookie.swift
```diff
@@ -0,0 +1,88 @@
...
+        properties[HTTPCookiePropertyKey.path] = path
+        properties[HTTPCookiePropertyKey.secure] = isSecure
+        properties[HTTPCookiePropertyKey.expires] = expiresDate
+        properties[HTTPCookiePropertyKey.comment] = comment
+        properties[HTTPCookiePropertyKey.commentURL] = commentURL
+        properties[HTTPCookiePropertyKey.discard] = isSessionOnly
+        properties[HTTPCookiePropertyKey.maximumAge] = expiresDate
+        properties[HTTPCookiePropertyKey.port] = portList
+        properties[HTTPCookiePropertyKey("HttpOnly")] = isHTTPOnly
+        if #available(iOS 13.0, *) {
```
lsanches, line 51: Same as above comment

lsanches (commented):
Left some comments. Otherwise looks good to me

FRAuth/FRAuth/Network/FRHTTPCookie.swift
george-bafaloukas-forgerock: This class is only used in the iOS 11+ scenario. But this is a good shout as I might need to specify some compiler flags here. Will have a look.

iOS 10 will always use HTTPCookie instead of the FRHTTPCookie

FRAuth/FRAuth/Network/FRHTTPCookie.swift
george-bafaloukas-forgerock: This class is only used in the iOS 11+ scenario. But this is a good shout as I might need to specify some compiler flags here. Will have a look.

iOS 10 will always use HTTPCookie instead of the FRHTTPCookie

FRAuth/FRAuth/Network/FRHTTPCookie.swift
george-bafaloukas-forgerock: Addressed. In other news iOS 10 simulators are no longer compatible with XCode and Big Sur, so no way to actually test that locally, but this class is now available in iOS 11+ environments

rodrigoareis (approved):
Changes looks good to me

spetrov (changes requested):
Please, put an entry in the changelog about this change.

george-bafaloukas-forgerock:
PR Updated please re - review

FRAuth/FRAuthTests/FRAuthSwiftTests/FRAuth/Cookie/CookieTests.swift
```diff
@@ -212,4 +212,82 @@ class CookieTests: FRAuthBaseTest {
         }
         XCTAssertEqual(count, 0)
     }
+    
+    func test_07_validate_cookie_is_stored_and_unarchived_Correctly() {
```
spetrov, line 216: 👍🏻

FRAuth/FRAuthTests/FRAuthSwiftTests/FRAuth/Cookie/CookieValidationTests.swift
```diff
@@ -202,7 +202,7 @@ class CookieValidationTests: FRAuthBaseTest {
     
     
     func test_10_cookie_is_expired_validation_not_expired() {
-        let setCookie: [String: String] = ["Set-Cookie":"iPlanetDirectoryPro=token; Expires=Wed, 21 Oct 2021 01:00:00 GMT; Domain=openam.example.com"]
+        let setCookie: [String: String] = ["Set-Cookie":"iPlanetDirectoryPro=token; Expires=Wed, 21 Oct 2022 01:00:00 GMT; Domain=openam.example.com"]
```
spetrov, line 205: 👍🏻

spetrov (approved):
Looks good to me! 👍🏻

rodrigoareis (approved):
Looks good to me

spetrov (approved):

PR #114, opened 2022-01-12
SDKS-1256 update google sign in

george-bafaloukas-forgerock:
Update GoogleSign In dependency and add SPM support


spetrov (approved):
LGTM 👍🏻

PR #120, opened 2022-01-31
SSL_pinning_iOS_draft_implementation

george-bafaloukas-forgerock:
SSL_pinning_iOS_draft_implementation

Adding first draft of SSL pinning implementation
Changing base supported version of all targets to 10.3
Resolving GoogleSignIn from SPM for the FRGoggleSignIn project similar to FaceBook
Adding FRPinningHandler & FRSecurityConfiguration in FRCore
Updating FRAuthConfig with optional "forgerock_pins" key and value example


vahancouver (commented):
Overall looks good and is aligned with what I have read and learned in the articles and tutorials about SSL pinning in iOS.

However I am not sure why we are using `URLSessionTaskDelegate` instead of `URLSessionDelegate`. If `URLSessionDelegate` is to be used, this will eliminate some code also (please see my comments)

I'd like to discuss the name _"forgerock_pins"_ for the property.

SampleApps/FRExample/FRExample/Configs/FRAuthConfig.plist
```diff
@@ -2,6 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+    <key>forgerock_pins</key>
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: I agree 100% with you on that one, I think we should come up with a better name for both

FRCore/FRCore/Network/FRPinningHandler.swift
```diff
@@ -0,0 +1,58 @@
...
+//
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+
+import Foundation
+
+@objc
+public protocol FRPinningHandlerProtocol: URLSessionTaskDelegate {
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: Let's discuss this, as the SDK was already using the URLSessionTaskDelegate, hence I left it for compatibility. You might right though and this might not be needed

vahancouver (commented):
Overall looks good. We are almost there. 

The only thing that I would change, is the hierarchy of classes. I think the class that implements the SSL Pinning (in this case FRURLSessionHandler) should be inherited from a more generic class (in this case DefaultURLSessionHandler) and not visa versa. And the client developers would implement their custom Session handler class inheriting from the more generic class (in this case DefaultURLSessionHandler). Of course we would need to update the class names. I would change the name of more generic class's from DefaultURLSessionHandler to FRURLSessionHandler and change more concrete class's name from FRURLSessionHandler to something like FRURLSessionCertificatePinningHandler (or something shorter). 

I have implemented my suggested changes locally and it works beautifully. It also eliminated (frSecurityConfiguration: nil) argument in the default (more generic) class's initializer.

Let's discuss over a call. Happy to show it.

FRAuth/FRAuthTests/FRAuthSwiftTests/FRAuth/SSL Pinning/SSLPinningTests.swift
```diff
@@ -87,7 +87,7 @@ class SSLPinningTests: FRAuthBaseTest {
         //  Init SDK
         self.startSDK()
         
-        let request = Request(url: "https://apple.com", method: .GET)
+        let request = Request(url: "https://webhook.site/309b8710-635b-40d6-9d90-dfbabeef954e", method: .GET)
```
george-bafaloukas-forgerock, line 90: Is this request URL something that will always be there or is it ephemeral causing the tests to break in the future?

FRCore/FRCore/Network/FRSecurityConfiguration.swift
```diff
@@ -104,6 +104,22 @@ public class FRSecurityConfiguration: NSObject {
...
     }
     
+    /// Extracts public key from the certificate
+    private func publicKey(for certificate: SecCertificate) -> SecKey? {
+        if #available(iOS 12.0, *) {
+            return SecCertificateCopyKey(certificate)
+        } else if #available(iOS 10.3, *) {
+            return SecCertificateCopyPublicKey(certificate)
+        } else {
+            var possibleTrust: SecTrust?
```
george-bafaloukas-forgerock, line 114: I am guessing we need to add that cause on SPM we cannot define minor versions. What happens if SPM is configured for 10.0, but the frameworks specify 10.3? Will it allow to get the Package and the project not run? Or would you get an error when checking out? The reason I am asking is cause, it might be better to go to 11 all-together or force 10.3 to avoid code we won't be able to test.

To my next point do we have a way to test the 10.0?

george-bafaloukas-forgerock: @spetrov what do you think of this regarding iOS 10.0?

FRAuth/FRAuthTests/FRAuthSwiftTests/FRAuth/SSL Pinning/SSLPinningTests.swift
vahancouver: Yes you are right. This was a tempoarary URL that would expire in 7 days. Changed to a permanent one

FRCore/FRCore/Network/FRSecurityConfiguration.swift
vahancouver: @george-bafaloukas-forgerock when initially I imported SPM (configured .v10) with frameworks that had iOS version set 10.3, it didn't build and and gave this error "'SecCertificateCopyPublicKey' is only available in iOS 10.3 or newer" and suggested to check for #availability. 

Seems like SPM doesn't look at framework versions, only what's set in the swift.package file. I tried this by setting my test project deployment target to 10.0, while the frameworks versions were set 10.3 and it built and run perfectly. So it only cared about the version set in SPM.

I also checked iOS 10 install base statistics, it's less than 0.5% . I am for moving to iOS 11.0 everywhere.

george-bafaloukas-forgerock: OK let's discuss this with @lsanches, @spetrov, Giessel and Bela. I am for going iOS 11 and save the hassle

vahancouver (approved):
After so many iterations, this looks complete. iOS version has been bumped to 11.0

spetrov (approved):
LGTM 👍🏻

PR #121, opened 2022-03-02
Sdks 1553 fix wrong version in logs for spm

vahancouver:
[PLEASE DO NOT MERGE THIS TO DEVELOP, THIS HAS DRAFT COMMITS THAT I WOULD NOT WANT TO BE PART OF THE TREE. I WILL HAVE A CLEANER BRANCH AFTER DISCUSSIONS]

So the problem we had was that we were reading the SDK bundle version from CFBundleShortVersionString in Info.plist. This was working great with cocoapods as cocoapods creates a project for each module under the main workspace. In case of SPM, we have only one project and it gets the main app's version (which is not what we want).

I have tried a few things, like adding a variable to Package.swift and using it as a value in target's build settings (I followed the way google did in the signin package) and then trying to read it in Info.plist. But that didn't work at all. 

I have also just tried to hardcode it in the Info.plist in each module, but then realized that we don't include the Info.plist files in the SPM package (we even explicitly exclude them). This also could have been the reason of my previous step not working at all.

One other version the internet was suggesting, was trying to read Package.resolved file and extract the version from there (even if I managed to do so, it wouldn't work if the package was imported by a branch name and not by a version).

So I have decided to hardcode the version as a static let in FRCore.Log class (in one place only) and use that value in all the bundles where we print the the SDK version. I know this is not ideal and requires an update with every release.


rodrigoareis (approved):
Changes looks good, but this is definitely not the ideal solution. Hopefully, a better solution will be provided in the future. I suggest to update the release steps to include it once we merge this PR.

PR #122, opened 2022-03-03
Add sdkVersion constant to FRCore.Log class and use it instead of CFBundleShortVersionString

vahancouver:
So the problem we had was that we were reading the SDK bundle version from CFBundleShortVersionString in Info.plist. This was working great with cocoapods as cocoapods creates a project for each module under the main workspace. In case of SPM, we have only one project and it gets the main app's version (which is not what we want).

I have tried a few things, like adding a variable to Package.swift and using it as a value in target's build settings (I followed the way google did in the signin package) and then trying to read it in Info.plist. But that didn't work at all.

I have also just tried to hardcode it in the Info.plist in each module, but then realized that we don't include the Info.plist files in the SPM package (we even explicitly exclude them). This also could have been the reason of my previous step not working at all.

One other version the internet was suggesting, was trying to read Package.resolved file and extract the version from there (even if I managed to do so, it wouldn't work if the package was imported by a branch name and not by a version).

So I have decided to hardcode the version as a static let in FRCore.Log class (in one place only) and use that value in all the bundles where we print the the SDK version. I know this is not ideal and requires an update with every release.

### DIFF
--- a/FRAuth/FRAuth/Log/FRLog.swift
+++ b/FRAuth/FRAuth/Log/FRLog.swift
@@ -23,11 +23,7 @@ By default, FRLog uses OSLog to display the log entry in the debug console, and 
     /// Module name of FRLog
     static var ModuleName: String {
         get {
-            var versionStr = ""
-            if let version = Bundle(for: FRAuth.self).infoDictionary?["CFBundleShortVersionString"] as? String {
-                versionStr = "[\(version)]"
-            }
-            return "[FRAuth]" + versionStr
+            return "[FRAuth]" + "[\(FRCore.Log.sdkVersion)]"
         }
     }
     
--- a/FRAuth/FRAuth/Log/FRLog.swift
+++ b/FRAuth/FRAuth/Log/FRLog.swift
@@ -2,7 +2,7 @@
 //  FRLog.swift
 //  FRAuth
 //
-//  Copyright (c) 2020-2021 ForgeRock. All rights reserved.
+//  Copyright (c) 2020-2022 ForgeRock. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
--- a/FRAuthenticator/FRAuthenticator/Log/FRALog.swift
+++ b/FRAuthenticator/FRAuthenticator/Log/FRALog.swift
@@ -23,11 +23,7 @@ public struct FRALog {
     /// Module name of FRLog
     static var ModuleName: String {
         get {
-            var versionStr = ""
-            if let version = Bundle(for: FRAClient.self).infoDictionary?["CFBundleShortVersionString"] as? String {
-                versionStr = "[\(version)]"
-            }
-            return "[FRAuthenticator]" + versionStr
+            return "[FRAuthenticator]" + "[\(FRCore.Log.sdkVersion)]"
         }
     }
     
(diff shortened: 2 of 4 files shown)


FRAuth/FRAuth/Log/FRLog.swift
```diff
@@ -23,11 +23,7 @@ By default, FRLog uses OSLog to display the log entry in the debug console, and
     /// Module name of FRLog
     static var ModuleName: String {
         get {
-            var versionStr = ""
-            if let version = Bundle(for: FRAuth.self).infoDictionary?["CFBundleShortVersionString"] as? String {
-                versionStr = "[\(version)]"
-            }
-            return "[FRAuth]" + versionStr
+            return "[FRAuth]" + "[\(FRCore.Log.sdkVersion)]"
```
rodrigoareis, line 26: Update Copyright header

FRAuthenticator/FRAuthenticator/Log/FRALog.swift
```diff
@@ -23,11 +23,7 @@ public struct FRALog {
     /// Module name of FRLog
     static var ModuleName: String {
         get {
-            var versionStr = ""
-            if let version = Bundle(for: FRAClient.self).infoDictionary?["CFBundleShortVersionString"] as? String {
-                versionStr = "[\(version)]"
-            }
-            return "[FRAuthenticator]" + versionStr
+            return "[FRAuthenticator]" + "[\(FRCore.Log.sdkVersion)]"
```
rodrigoareis, line 26: Update Copyright header

FRCore/FRCore/Log/Log.swift
```diff
@@ -128,6 +128,8 @@ public class Log: NSObject {
     
     //  MARK: - Property
     
+    /// Current SDK version. We hard code it here as currently there is no other way to get it dinamically when used with SPM
```
rodrigoareis, line 131: Update Copyright header

FRProximity/FRProximity/Log/FRPLog.swift
```diff
@@ -22,11 +22,7 @@ public struct FRPLog {
     /// Module name of FRLog
     static var ModuleName: String {
         get {
-            var versionStr = ""
-            if let version = Bundle(for: FRProximity.self).infoDictionary?["CFBundleShortVersionString"] as? String {
-                versionStr = "[\(version)]"
-            }
-            return "[FRProximity]" + versionStr
+            return "[FRProximity]" + "[\(FRCore.Log.sdkVersion)]"
```
rodrigoareis, line 25: Update Copyright header

rodrigoareis (changes requested):
Changes looks good to me, missing copyright header updates

FRAuth/FRAuth/Log/FRLog.swift
vahancouver: done ✅

FRAuthenticator/FRAuthenticator/Log/FRALog.swift
vahancouver: done ✅

FRCore/FRCore/Log/Log.swift
vahancouver: done ✅

FRProximity/FRProximity/Log/FRPLog.swift
vahancouver: done ✅

rodrigoareis (approved):
Looks good to me

spetrov (approved):
Looks good to me! 👍🏻

PR #123, opened 2022-03-03
SDKS-1632: Sign in With Apple Full name updates

george-bafaloukas-forgerock:
- Added Helper parsing structs for Sign In With Apple Response Object
- Refactored 'AppleSignInHandler' to return the new JSON type of token, with the Parsed JSON response body
- Fixes for the assignment of Scopes for Sign in With Apple


spetrov (changes requested):
Changes look good to me. I have a few comments to be considered/fixed...
In addition, can you please update the `CHANGELOG.md` with a placeholder for the next release and list this change in there... (otherwise, we may forget it later...) For example in Android SDK we have this:

> ## [Unreleased]
> #### Added
> - Support SSL Pinning [SDKS-80]

FRAuth/FRAuth/Callbacks/IdPCallback.swift
```diff
@@ -101,7 +105,7 @@ public class IdPCallback: MultipleValuesCallback {
             }
         }
 
-        self.idpClient = IdPClient(provider: provider, clientId: clientId, redirectUri: redirect_uri, scopes: scopeValues, nonce: nonceValue, acrValues: acrValues, request: requestValue, requestUri: requestUriValue)
+        self.idpClient = IdPClient(provider: provider, clientId: clientId, redirectUri: redirect_uri, scopes: scopeValues, nonce: nonceValue, acrValues: acrValues, request: requestValue, requestUri: requestUriValue, acceptsJSON: acceptsJSON ?? false)
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: I will confirm if that breaks anything else and update the tests as needed

FRAuth/FRAuth/SocialLogin/AppleSignInUserStructs.swift
```diff
@@ -0,0 +1,79 @@
...
+//  This software may be modified and distributed under the terms
+//  of the MIT license. See the LICENSE file for details.
+//
+
+
+import Foundation
+import AuthenticationServices
+
+/*
+ Apple Sign in example JSON response for 3rd party applications.
```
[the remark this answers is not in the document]
george-bafaloukas-forgerock: Will add some tests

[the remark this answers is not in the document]
george-bafaloukas-forgerock: I am waiting to finilise the change basically and then will add the tests

FRAuth/FRAuth/SocialLogin/AppleSignInHandler.swift
```diff
@@ -115,12 +125,19 @@ extension AppleSignInHandler: ASAuthorizationControllerDelegate {
...
                 return
             }
             
-            self.completionCallback?(idToken, self.tokenType, nil)
+            if self.acceptsJSON == true {
+                self.completionCallback?(IDToken1token, self.tokenType, nil)
+            } else {
+                self.completionCallback?(id_token, self.tokenType, nil)
+            }
+
```
jeyanthanperiyasamy, line 140: either way works , but i am thinking of like this.   self.completionCallback?(self.acceptsJSON == true ? IDToken1token : id_token , self.tokenType, nil)

FRAuth/FRAuth/SocialLogin/AppleSignInUserStructs.swift
```diff
@@ -0,0 +1,79 @@
...
+     "user": {
+         "name": {
+             "firstName": "Kurt",
+             "lastName": "Cobain"
+         },
+         "email": "kurt@example.com"
+     }
+ }
+         
+ */
```
jeyanthanperiyasamy, line 31: Do we want test for this class ? i am not usually super worried if its just a getter and setter , but this has some logic for code , id_token, email

george-bafaloukas-forgerock: yeap I think some tests around this would be good. I am working to add some stuff on testing

FRAuth/FRAuth/SocialLogin/AppleSignInHandler.swift
george-bafaloukas-forgerock: I personally prefer the verbosity of the existing solution as it makes it easier to test

FRAuth/FRAuth/SocialLogin/AppleSignInHandler.swift
```diff
@@ -115,12 +125,19 @@ extension AppleSignInHandler: ASAuthorizationControllerDelegate {
         switch authorization.credential {
         case let appleIDCredential as ASAuthorizationAppleIDCredential:
             FRLog.v("ASAuthorizationAppleIDCredential received: \(appleIDCredential)")
-            guard let tokenData = appleIDCredential.identityToken, let idToken = String(data: tokenData, encoding: .utf8) else {
+                      
+            let appleSignInResponse = AppleSignInResponse(appleIDCredential)
+            guard let authorization_code = appleSignInResponse.code, let id_token = appleSignInResponse.id_token, let IDToken1tokenJSON = try? JSONEncoder().encode(appleSignInResponse), let IDToken1token = String(data: IDToken1tokenJSON, encoding: .utf8) else {
                 self.completionCallback?(nil, nil, SocialLoginError.unsupportedCredentials("Failed to parse received credentials data (ASAuthorizationAppleIDCredential.identityToken)"))
                 return
             }
```
vahancouver, line 134: What are the chances that the above code might result in an early return when _acceptsJSON = false_ and we only need the _id_token_ ? I'm just making sure we don't break anything for the older versions here by adding all this checks and trying to create a JSON when we don't really need it.

vahancouver, line 134: What are the chances that the above code might result in an early return when _acceptsJSON = false_ and we only need the _id_token_ ? I'm just making sure we don't break anything for the older versions here by adding all this checks and trying to create a JSON when we don't really need it.

vahancouver (commented):
Overall looks really good. 
Just a comment and a question.

FRAuth/FRAuth/SocialLogin/AppleSignInHandler.swift
george-bafaloukas-forgerock: This shouldn't happen as apple returns all the fields, but to make sure and not have unnecessary assignments that we don't need I split it. Good observation

spetrov (approved):

witrisna (approved):
Looks good to me.

rodrigoareis (approved):
Changes looks good to me

PR #125, opened 2022-04-05
Sdks-1708_inject cookie to header instead of query param

george-bafaloukas-forgerock:
Fix in OAuth2Client class to pass the cookie(sso token) as a header field, not a query param


vahancouver (approved):
LGTM

spetrov (approved):
LGTM 👍🏻

PR #128, opened 2022-04-25
Sdks 1491 expired push notification displayed as approved

vahancouver:
Fix the bug when expired notification would show up as approved
Move saving notification to the storage to the request success block (similar to Android) - this also fixed the failing tests after the first commit

### DIFF
--- a/FRAuthenticator/FRAuthenticator/Model/Notification/PushNotification.swift
+++ b/FRAuthenticator/FRAuthenticator/Model/Notification/PushNotification.swift
@@ -2,7 +2,7 @@
 //  Notification.swift
 //  FRAuthenticator
 //
-//  Copyright (c) 2020-2021 ForgeRock. All rights reserved.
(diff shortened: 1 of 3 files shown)


FRAuthenticator/FRAuthenticator/Model/Notification/PushNotification.swift
```diff
@@ -240,40 +240,40 @@ public class PushNotification: NSObject, NSSecureCoding, Codable {
     ///   - result: Boolean indicator whether or not PushNotification authentication is approved or denied
```
spetrov, line 240: Since the parameter name was changed, please update the name in the inline docs as well (i.e. "result" -> "approved")

spetrov (changes requested):
Looks good to me! 
I have only one minor comment...

spetrov (approved):
👍🏻

rodrigoareis (approved):
Changes looks good to me

PR #129, opened 2022-04-26
SDKS-1665 - Obtain timestamp from new Push Notification payload

vahancouver:
Set _timeAdded_ from the new payload if available. Update and add unit tests

Please note that in iOS PushNotification doesn't have _timeExpired_ property. It has _isExpired_ property that is calculated based on _timeAdded_ and _ttl_

### DIFF
--- a/FRAuthenticator/FRAuthenticator/Model/Notification/PushNotification.swift
+++ b/FRAuthenticator/FRAuthenticator/Model/Notification/PushNotification.swift
@@ -117,7 +117,11 @@ public class PushNotification: NSObject, NSSecureCoding, Codable {
         self.ttl = ttlDouble
         self.mechanismUUID = mechanismUUID
         
-        self.timeAdded = Date()
+        if let intervalString = payload["i"] as? String, let interval = Int64(intervalString) {
+            self.timeAdded = Date(milliseconds: interval)
+        } else {
+            self.timeAdded = Date()
+        }
     }
     
     
--- a/FRAuthenticator/FRAuthenticatorTests/UnitTests/Model/NotificationTests.swift
+++ b/FRAuthenticator/FRAuthenticatorTests/UnitTests/Model/NotificationTests.swift
@@ -2,7 +2,7 @@
 //  NotificationTests.swift
 //  FRAuthenticatorTests
 //
-//  Copyright (c) 2020-2021 ForgeRock. All rights reserved.
+//  Copyright (c) 2020-2022 ForgeRock. All rights reserved.
 //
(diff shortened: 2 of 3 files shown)


FRAuthenticator/FRAuthenticator/Model/Notification/PushNotification.swift
```diff
@@ -117,7 +117,11 @@ public class PushNotification: NSObject, NSSecureCoding, Codable {
         self.ttl = ttlDouble
         self.mechanismUUID = mechanismUUID
         
-        self.timeAdded = Date()
+        if let intervalString = payload["i"] as? String, let interval = Int64(intervalString) {
+            self.timeAdded = Date(milliseconds: interval)
```
rodrigoareis, line 121: This needs to use: Date(timeIntervalSince1970:)

FRAuthenticator/FRAuthenticatorTests/UnitTests/Model/NotificationTests.swift
```diff
@@ -28,6 +29,7 @@ class NotificationTests: FRABaseTests {
             XCTAssertEqual(notification.loadBalanceKey, "amlbcookie=01")
             XCTAssertEqual(notification.ttl, t)
             XCTAssertEqual(notification.mechanismUUID, u)
+            XCTAssertEqual(notification.timeAdded, Date(milliseconds: i))
```
rodrigoareis, line 32: Tests also should use Date(timeIntervalSince1970:)

rodrigoareis (changes requested):
Changes looks good to me. But I think we need to use Date(timeIntervalSince1970:) to parse the timestamp. I remember it was required to use it to obtain the correct Date while updating the serialization methods for the PushNotification class.

FRAuthenticator/FRAuthenticator/Model/Notification/PushNotification.swift
rodrigoareis: The code is already using the Date extension that parses the date as Date(timeIntervalSince1970:). No need to change.

FRAuthenticator/FRAuthenticatorTests/UnitTests/Model/NotificationTests.swift
rodrigoareis: Same as above.

rodrigoareis (approved):
All good

spetrov (approved):
LGTM 👍🏻

PR #132, opened 2022-05-05
SDKS-1826 Remove the example item from forgerock_ssl_pinning_public_key_hashes 

vahancouver:
Remove the example item from forgerock_ssl_pinning_public_key_hashes  array in the sample app's plist


spetrov (approved):
👍🏻

PR #156, opened 2022-09-07
Added workflow for triggering the e2e tests against IDCloud instance.

spetrov:
# JIRA Ticket

[SDKS-1789](https://bugster.forgerock.org/jira/browse/SDKS-1789) "Automate Cloud Upgrade Testing"

# Description

This PR adds workflow for triggering the e2e test cases against live IDCloud environment. The workflow is called "Run Live Tests". A new `FRTestHostE2E.xcscheme` scheme has been added to the project to filter the "live" test cases only.
The e2e tests now can be triggered either manually from the "Actions" menu in the repo, or via an HTTP request. 
Triggering the workflow via HTTP request requires a [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) (PAT) passed as a Bearer token in the Authorization header. For more details read the [workflow-dispatch-event](https://docs.github.com/en/rest/actions/workflows#create-a-workflow-dispatch-event) docs.

### DIFF
--- a/FRTestHost/FRTestHost/SharedTestFiles/TestConfig/Config-live-01.json
+++ b/FRTestHost/FRTestHost/SharedTestFiles/TestConfig/Config-live-01.json
@@ -1,1 +1,52 @@
-
+{
+    "username": "sdkuser",
+    "password": "*******",
+    "kba": {
+        "What's your favorite color?": "White",
+        "Who was your first employer?": "ForgeRock"
+    },
+    "user-email": "sdkuser@example.com",
+    "user-first-name": "Sdk",
+    "user-last-name": "User",
+    "user-info": {
+            "given_name": "Sdk",
+            "family_name": "User",
+            "middle_name": "",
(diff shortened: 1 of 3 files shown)


FRTestHost/FRTestHost/SharedTestFiles/TestConfig/Config-live-01.json
```diff
@@ -1 +1,52 @@
-
+{
+    "username": "sdkuser",
+    "password": "password",
```
rodrigoareis, line 3: Is it ok to expose the password?

spetrov: uh, oh! of course not! 
I fixed that by putting the password in github actions secret...

rodrigoareis (approved):
Changes looks good to me

PR #157, opened 2022-09-15
Add verification for State value when calling the Authorize endpoint

vahancouver:
# JIRA Ticket

[SDKS-2077](https://bugster.forgerock.org/jira/browse/SDKS-2077)

# Description

Add verification for State value when calling the Authorize endpoint

### DIFF
--- a/FRAuth/FRAuth/Config/OAuth2Client.swift
+++ b/FRAuth/FRAuth/Config/OAuth2Client.swift
@@ -208,11 +208,12 @@ public class OAuth2Client: NSObject, Codable {
                     let redirectURL = URL(string: redirectURLAsString)
                     
                     // Verify that the response state value is the same
-                    if let state = redirectURL?.valueOf("state"), state != pkce.state {
+                    guard let state = redirectURL?.valueOf("state"), state == pkce.state else {
                         completion(nil, OAuth2Error.invalidPKCEState)
+                        return
                     }
                     //  If authorization_code was included in the redirecting request, extract the code, and continue with token endpoint
-                    else if let authCode = redirectURL?.valueOf("code") {
+                    if let authCode = redirectURL?.valueOf("code") {
                         
                         let request = self.buildTokenWithCodeRequest(code: authCode, pkce: pkce)
                         FRRestClient.invoke(request: request, action: Action(type: .EXCHANGE_TOKEN), completion: { (result) in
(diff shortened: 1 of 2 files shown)


rodrigoareis (approved):
Change looks good to me

cerebrl (approved):
The changes look right to me. I have some feedback on this code, but I'm just suggesting this for future consideration, _not_ as part of this PR.

The code in the async and sync functions look really similar if not exactly the same. Is there anything that stops us from pulling that code out and converting it into a reusable function?

There's also quite a bit of if-else-if-else statements, I _think_ this could be cleaned up just a touch by pulling all error conditions up in the function and returning early (that should be a thing in Swift, right?). So, that way, exit the function early and get the error conditions out of the way, and then the rest of the function is reserved for the success logic, no wrapping of conditions necessary.

As it is now, we have error conditions and success conditions all mixed together. Anyway, just my two cents.

FRAuth/FRAuth/Config/OAuth2Client.swift
```diff
@@ -207,8 +207,12 @@ public class OAuth2Client: NSObject, Codable {
                     
                     let redirectURL = URL(string: redirectURLAsString)
                     
+                    // Verify that the response state value is the same
+                    if let state = redirectURL?.valueOf("state"), state != pkce.state {
```
spetrov, line 211: It is weird that `state` is part of the `pkce` object.
I don't think we should address this in this PR since it has been like that for a very long time.
The `state` is part of the [OAuth2 spec](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2), and PKCE is another [spec](https://datatracker.ietf.org/doc/html/rfc7636)... 
Let's discuss this on Friday's meeting...

spetrov, line 211: It looks to me that when the `state` param is omitted by the server the client won't throw an error... 
We should fix that because this is the whole purpose of `state`... reference: [OAuth2 spec](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2)

FRAuth/FRAuth/Config/OAuth2Client.swift
```diff
@@ -291,8 +295,12 @@ public class OAuth2Client: NSObject, Codable {
                 
                 let redirectURL = URL(string: redirectURLAsString)
                 
+                // Verify that the response state value is the same
+                if let state = redirectURL?.valueOf("state"), state != pkce.state {
```
spetrov, line 299: Same here...

spetrov (changes requested):
Please, add some test cases covering cases where the `state` in the response is present, but different from the initial one... and also when `state` is completely missing in the response...

FRAuth/FRAuth/Config/OAuth2Client.swift
vahancouver: Addressed the comment.

FRAuth/FRAuth/Config/OAuth2Client.swift
vahancouver: Addressed the comment.

spetrov (approved):
Looks good to me! 👍🏻

PR #161, opened 2022-10-13
Update links for new docs layout.

forgerock-chris:
SDK documentation now exists under `/latest`, and BackStage is our preferred URL base.


spetrov (approved):
Looks good! 👍🏻

PR #165, opened 2022-11-01
SDKS-1750 Device Binding e2e test cases

spetrov:
# JIRA Ticket

[SDKS-1750](https://bugster.forgerock.org/jira/browse/SDKS-1750) "Device Binding QA"

# Description
- e2e test cases for the Device Binding feature.
- These test cases assume that the target IDCloud instance supports the DeviceBinding node and contains the following authentication tree: 

image


vahancouver (approved):
LGTM

PR #171, opened 2022-11-22
Adjusting error case handling for WAKError

denissb:
# Better error handling on the iOS platform

Right now there is not distinction between different types of LAError returned by the iOS platform.
Without this we can't make a distinction based on the type of error in our flows.

This is a proposal, I don't know the best way to map the WebAuthn error codes.


spetrov:
Closing per Jey's comment above.
We addressed this issue with [SDKS-2226](https://bugster.forgerock.org/jira/browse/SDKS-2226) "Added error message description to the `WebAuthnError` enum" (#172)

PR #184, opened 2023-01-25
Fix a memory leak in DefaultPinCollector

vahancouver:
# JIRA Ticket

[SDKS-2237](https://bugster.forgerock.org/jira/browse/SDKS-2237)

# Description

Remove the unnecessary code from _DispatchQueue.main.async_ closure
Fix a memory leak caused by a retain cycle


jeyanthanperiyasamy (approved):

rodrigoareis (approved):
Looks good to me

PR #199, opened 2023-04-17
Move Device Binding to a separate module

vahancouver:
# JIRA Ticket

[SDKS-2245](https://bugster.forgerock.org/jira/browse/SDKS-2245) iOS - DeviceBind in separate module

# Description

Move Device Binding code and tests to a new FRDeviceBinding module

1. Please pay attention to how I removed **DeviceBindingCallback** and **DeviceSigningVerifierCallback** from _CallbackFactory.swift_ and now have to register these two new callbacks after SDK start (like in _SampleApps/FRExample/FRExample/ViewController.swift_ ). Ideally we want to find a way to avoid being have to do this)
2. A few classes/methods are now public in FRAuth to be accessible from FRDeviceBinding module
3. _RemoteDeviceBindingRepository_ chnaged to use _options_ instead of _serverConfig_ as it is already public and contains the info we need


FRAuth/FRAuth/Callbacks/CallbackFactory.swift
```diff
@@ -81,9 +81,7 @@ public class CallbackFactory: NSObject {
             CallbackType.WebAuthnRegistrationCallback.rawValue: WebAuthnRegistrationCallback.self,
             CallbackType.IdPCallback.rawValue: IdPCallback.self,
             CallbackType.SelectIdPCallback.rawValue: SelectIdPCallback.self,
-            CallbackType.DeviceBindingCallback.rawValue: DeviceBindingCallback.self,
-            CallbackType.DeviceSigningVerifierCallback.rawValue: DeviceSigningVerifierCallback.self
-        ]
+        ]
```
jeyanthanperiyasamy, line 84: For add the callback dynamically, its going to be tricky because we cannot import FRDeviceBind module in FRAuth that would lead to a Circular dependency between modules. so far i can think of couple of ideas

1. So my first idea is in FRUSerkey class , which is public facing , but this can be changed and let the user to call the register from their original application, which is kind of work around 

class FRUSerkey {

 public func register() {
        CallbackFactory.shared.registerCallback(callbackType: "DeviceBindingCallback", callbackClass: DeviceBindingCallback.self)
        CallbackFactory.shared.registerCallback(callbackType: "DeviceSigningVerifierCallback", callbackClass: DeviceSigningVerifierCallback.self)
    }
}


2. This approach will work for sure , but will require lot of work.
How about DeviceBind module to just depend only on FRCore Module? so that we can fix the circular dependency issue. and we can easily do this

#if canImport(FRDeviceBinding)
    import FRDeviceBinding
#endif
 


This is the easiest approach , How about we add a method in FRDeviceBind Module to register

jeyanthanperiyasamy:
middle of reviewing, will continue this tomorrow

FRAuth/FRAuth/Callbacks/CallbackFactory.swift
rodrigoareis, line 84: Update copyright header

rodrigoareis: +1

FRAuth/FRAuth/Constants/CallbackConstants.swift
```diff
@@ -96,7 +96,7 @@ enum CallbackType: String {
 }
 
 /// CBConstants is mainly responsible to maintain all constant values related to Callback implementation
-struct CBConstants {
+public struct CBConstants {
```
rodrigoareis, line 99: Update copyright header

FRAuth/FRAuth/Device/FRDevice.swift
```diff
@@ -16,7 +16,7 @@ import Foundation
     /// Static FRDeivce instance for singleton
     static var _staticDevice: FRDevice? = nil
     /// FRDeviceIdentifier object responsible for device's instance identifier
-    fileprivate(set) var identifier: FRDeviceIdentifier
+    public fileprivate(set) var identifier: FRDeviceIdentifier
```
rodrigoareis, line 19: Update copyright header

FRAuth/FRAuthTests/FRAuthSwiftTests/E2ETests/Callback-Live/AA-05-DeviceBindingCallbackTest.swift
```diff
@@ -10,6 +10,7 @@
 
 import XCTest
 @testable import FRAuth
+@testable import FRDeviceBinding
```
rodrigoareis, line 13: Update copyright header

FRDeviceBinding/FRDeviceBinding/Sources/CBConstants.swift
```diff
@@ -0,0 +1,42 @@
+//
```
rodrigoareis, line 1: Update copyright header

FRDeviceBinding/FRDeviceBindingTests/DeviceSigningVerifierCallbackTests.swift
```diff
@@ -12,6 +12,7 @@
 import XCTest
 @testable import FRAuth
 @testable import FRCore
+@testable import FRDeviceBinding
```
rodrigoareis, line 15: Update copyright header

rodrigoareis (changes requested):
There are some missing copyright update on changed files.

In my opinion, unless we can handle the registration of device binding callbacks dynamically, it's not worth have a separated module for device binding.

vahancouver:
> There are some missing copyright update on changed files.
> 
> In my opinion, unless we can handle the registration of device binding callbacks dynamically, it's not worth have a separated module for device binding.

@rodrigoareis updated the copyrights. Found some more places where I didn't update the module name in the header comments. Updated those too

FRAuth/FRAuth/Callbacks/CallbackFactory.swift
vahancouver: @jeyanthanperiyasamy I thought about the first approach. Probably the best is to create a new class with the same name as the module (FRDeviceBinding) and add the register method there. 

As for the second one, Yes you are right, I tried it with FRAuth and ran into circular dependency issue. I am not sure we will be able to import just FRCore, as FRDeviceBinding has so many dependencies from FRAuth. Let's connect and discuss this one.

jeyanthanperiyasamy (commented):
yeah overall looks good. Except lets discuss with team about adding the callbacks dynamically  .

FRAuth/FRAuth/Callbacks/CallbackFactory.swift
witrisna: Is there a way to check if the class or framework exists?
https://stackoverflow.com/questions/42099811/how-to-check-if-ios-application-uses-storekit-framework

FRAuth/FRAuth/FRAuth.swift
```diff
@@ -83,10 +83,19 @@ public final class FRAuth: NSObject {
             try FRAuth.initWithOptions(options: configOptions)
         }
         
+        // If FRProximity module is present, invoke FRProximity.startProximity()
```
rodrigoareis, line 86: Update copyright

rodrigoareis (approved):
Changes looks good to me

spetrov (approved):

PR #208, opened 2023-05-25
Add SwiftUI QuickStart Example

vahancouver:
# JIRA Ticket

[SDKS-2405](https://bugster.forgerock.org/jira/browse/SDKS-2405)

# Description

[Sample App] iOS - Swift UI sample App


SampleApps/QuickStart-SwiftUI/QuickStart/ViewModels/CallbackViewModel.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+// Copyright (c) 2023 ForgeRock. All rights reserved.
+//
+// This software may be modified and distributed under the terms
+// of the MIT license. See the LICENSE file for details.
+//
+
+import Foundation
+
+class CallbackViewModel: ObservableObject {
```
jeyanthanperiyasamy, line 13: do you need a ObservableObject here ? since this model is not publishing anything

SampleApps/QuickStart-SwiftUI/QuickStart/ViewModels/NodeViewModel.swift
```diff
@@ -0,0 +1,73 @@
...
+        }
+        
+        if callbackViewModels.count > 0 {
+            statusViewModel?.status = Status(statusDescription: "Node successfully parsed", statusType: .info)
+        }
+        
+    }
+    
+    func submitNode() {
+        guard let node = node, let callbacks = node.callbacks as? [SingleValueCallback] else { return }
```
george-bafaloukas-forgerock, line 51: I am not 100% sure about this. It works on this example, but the moment you add more callback types it will need refactoring

SampleApps/QuickStart-SwiftUI/QuickStart/ViewModels/NodeViewModel.swift
```diff
@@ -0,0 +1,73 @@
...
+    }
+    
+    func parseNode() {
+        guard let node = node else {
+            print("Node is missing")
+            return
+        }
+        
+        for callback in node.callbacks {
+            if callback.isKind(of: NameCallback.self), let nameCallBack = callback as? NameCallback {
```
george-bafaloukas-forgerock, line 34: Why do you check both `callback.isKind(of: NameCallback.self)` and `let nameCallBack = callback as? NameCallback`

The latter would be enough and simple as it does the casting as well

george-bafaloukas-forgerock (changes requested):
A couple minor changes needed, other than that looks great. Good job

SampleApps/QuickStart-SwiftUI/QuickStart/ViewModels/ContentViewModel.swift
```diff
@@ -0,0 +1,32 @@
...
+
+import Foundation
+import FRAuth
+
+@MainActor class ContentViewModel: ObservableObject {
+    @Published private(set) var sdkStarted = false
+    
+    func startSDK(statusViewModel: StatusViewModel) {
+        FRLog.setLogLevel([.network, .all])
+        FRAuth.configPlistFileName = "FRAuthConfig"
```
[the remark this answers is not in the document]
vahancouver: Removed the line and added comments about specifying the file name or using FROptions for dynamic configuration

SampleApps/QuickStart-SwiftUI/QuickStart/ViewModels/NodeViewModel.swift
vahancouver: We check earlier i the code for supported Callbacks and show error if we encounter a callback that is not supported. That being said I moved the callback type casting/checking down in the for loop. safer this way

george-bafaloukas-forgerock (approved):
nice one

spetrov (approved):
LGTM! 👍🏻

PR #220, opened 2023-06-30
SDKS-2563 - Making "refresh" and "refreshSync" methods Public

george-bafaloukas-forgerock:
- Making "refresh" and "refreshSync" methods Public
- Updating tests

# JIRA Ticket

Please, link jira ticket here.
SDKS-2563

### DIFF
--- a/FRAuth/FRAuth/User/FRUser.swift
+++ b/FRAuth/FRAuth/User/FRUser.swift
@@ -297,6 +297,47 @@ public class FRUser: NSObject, NSSecureCoding {
         }
     }
     
+    /// Refreshes user's session with refresh_token regardless of validity of current access_token
+    ///
+    /// - Parameter completion: Completion callback that notifies the result of operation
+    @objc
+    public func refresh(completion:@escaping UserCallback) {
+        if let frAuth = FRAuth.shared, let tokenManager = frAuth.tokenManager {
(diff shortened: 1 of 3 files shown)


jeyanthanperiyasamy (approved):

vahancouver (approved):
Changes look good

spetrov (changes requested):
@george-bafaloukas-forgerock, the changes look good to me. 
However, during testing, I realized that refreshing the access_token does not revoke the previous one. I thought AM would automatically invalidate it, but it only invalidates the refresh_token... 
It looks to me it is up to the client to revoke the old access_token, and we currently don't do that, so I would like to discuss this with you and the team before merging and releasing this change...

Here are the relevant parts from RFC 6749:
- [Section 1.5 Refresh Token](https://datatracker.ietf.org/doc/html/rfc6749#section-1.5)

> Refresh tokens are issued to the client by the authorization server and are
   used to obtain a new access token when the current access token
   becomes invalid or expires, **or to obtain additional access tokens**
   with identical or narrower scope

- [Section 6 Refreshing an Access Token](https://datatracker.ietf.org/doc/html/rfc6749#section-6)

> The authorization server **MAY** revoke the old
    refresh token after issuing a new refresh token to the client.

spetrov (approved):

PR #223, opened 2023-07-18
Revoke old access_token upon force refresh

vahancouver:
[SDKS-2571](https://bugster.forgerock.org/jira/browse/SDKS-2571)

Revoke old access_token upon force refresh. Made sure it's only revoking the access token and not the refresh token.

**The following regression test has been completed (this test has been verified with "Issue Refresh Tokens on Refreshing Access Tokens" both On and Off in AM console)**

1. Call Login - Verify that access token and refresh token has been issued

2. Call Get User Info - Verify that user info successfully retrieved using the Access Token from Step 1

3. Call Refresh Token - Verify that the new Access Token has been issued (check that it's different from the Access Token from Step 1) and that Revoke endpoint has been called with the old Access Token from Step 1

4. Call Get User Info - Verify that user info successfully retrieved using the Access Token from Step 3

5. In the Terminal call /userinfo endpoint using the Access Token from Step 1 - Verify that "invalid_token" response has been received

### DIFF
--- a/FRAuth/FRAuth/Manager/TokenManager.swift
+++ b/FRAuth/FRAuth/Manager/TokenManager.swift
@@ -2,7 +2,7 @@
 //  TokenManager.swift
 //  FRAuth
 //
-//  Copyright (c) 2019-2022 ForgeRock. All rights reserved.
+//  Copyright (c) 2019-2023 ForgeRock. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
@@ -364,5 +364,13 @@ struct TokenManager {
         FRSession._staticSession = nil
         FRUser._staticUser = nil
         Browser.currentBrowser = nil
+    }
+    
+    
+    /// Revoke given token
+    func revokeToken(_ token: AccessToken, completion: @escaping CompletionCallback) {
+        self.oAuth2Client.revoke(accessToken: token) { (error) in
+            completion(error)
+        }
     }
 }
--- a/FRAuth/FRAuth/Manager/TokenManager.swift
+++ b/FRAuth/FRAuth/Manager/TokenManager.swift
@@ -367,9 +367,9 @@ struct TokenManager {
     }
     
     
-    /// Revoke given token
+    /// Revoke given Access Token without using the Refresh Token
     func revokeToken(_ token: AccessToken, completion: @escaping CompletionCallback) {
-        self.oAuth2Client.revoke(accessToken: token) { (error) in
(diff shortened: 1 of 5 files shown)


jeyanthanperiyasamy:
Can you do a regression test like this by mocking all the endpoints ? 

1. GetAccessToken
2. Call RefreshToken
3. verify the revoke endpoint called with the previous token
4.  verify the previous token is not available in the cache

jeyanthanperiyasamy (commented):
comments added

FRAuth/FRAuth/Manager/TokenManager.swift
```diff
@@ -365,4 +365,12 @@ struct TokenManager {
         FRUser._staticUser = nil
         Browser.currentBrowser = nil
     }
+    
+    
+    /// Revoke given token
+    func revokeToken(_ token: AccessToken, completion: @escaping CompletionCallback) {
+        self.oAuth2Client.revoke(accessToken: token) { (error) in
```
george-bafaloukas-forgerock, line 372: @vahancouver I think this is not going to work.
Calling this will call:
`public func revoke(accessToken: AccessToken, completion: @escaping CompletionCallback) {
        // Construct parameter for the request
        var parameter:[String: String] = [:]
        let token = accessToken.refreshToken ?? accessToken.value
        parameter[OAuth2.token] = token
        parameter[OAuth2.clientId] = self.clientId
        
        let request = Request(url: self.serverConfig.tokenRevokeURL, method: .POST, headers: [:], bodyParams: parameter, urlParams: [:], requestType: .urlEncoded, responseType: .json, timeoutInterval: self.serverConfig.timeout)
        
        FRRestClient.invoke(request: request, action: Action(type: .REVOKE_TOKEN)) { (result) in
            switch result {
            case .success(_ , _):
                completion(nil)
                break
            case .failure(let error):
                completion(error)
                break
            }
        }
    }
    `

george-bafaloukas-forgerock: That code will revoke based on the Refresh token, by default and will thus revoke all accessTokens created by this refresh token. And the refresh token as well. Unless mistaken, this will end up saving a token that has been revoked

george-bafaloukas-forgerock (changes requested):

george-bafaloukas-forgerock:
As a general note, I would think we would need to change the `OAuth2Client` revoke method to take an input whether to revoke using the refresh token or the accessToken

george-bafaloukas-forgerock (approved):
Looks good. Make sure you test the edge cases with the Refresh tokens please

spetrov (approved):
LGTM! 👍🏻

PR #233, opened 2023-09-21
SafariviewController reader fix

george-bafaloukas-forgerock:
Setting the `config.entersReaderIfAvailable` to "false" as this can c…ause the centralized login flow to load in a reader view blocking users to authenticate

# JIRA Ticket

SDKS-2746

# Description

### DIFF
--- a/FRAuth/FRAuth/User/Browser.swift
+++ b/FRAuth/FRAuth/User/Browser.swift
@@ -316,7 +316,7 @@ import SafariServices
         var viewController: SFSafariViewController?
         if #available(iOS 11.0, *) {
             let config = SFSafariViewController.Configuration()
-            config.entersReaderIfAvailable = true
+            config.entersReaderIfAvailable = false
             viewController = SFSafariViewController(url: url, configuration: config)
             viewController?.delegate = self
         }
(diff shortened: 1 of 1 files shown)


FRAuth/FRAuth/User/Browser.swift
```diff
@@ -316,7 +316,7 @@ import SafariServices
         var viewController: SFSafariViewController?
         if #available(iOS 11.0, *) {
             let config = SFSafariViewController.Configuration()
-            config.entersReaderIfAvailable = true
+            config.entersReaderIfAvailable = false
             viewController = SFSafariViewController(url: url, configuration: config)
             viewController?.delegate = self
         }
```
vahancouver, line 322: false is the default value. I wonder if we should still explicitly set it. Also as we are changing, maybe we should get rid of the iOS 11 check as we support iOS 12+. So I suggest to replace lines 317-326 with the following:

        viewController = SFSafariViewController(url: url, configuration: SFSafariViewController.Configuration())
        viewController?.delegate = self

vahancouver (commented):
Left a comment

vahancouver (approved):
LGTM

jeyanthanperiyasamy (approved):

rodrigoareis (approved):
Changes looks good to me

spetrov (approved):
LGTM! 👍🏻

PR #235, opened 2023-09-22
Set default Session Configuration's urlCache to nil

vahancouver:
[SDKS-2700](https://bugster.forgerock.org/jira/browse/SDKS-2700)
[iOS] Inadequate Cache Control - Mobile Application

Approach 2

### DIFF
--- a/FRCore/FRCore/Network/RestClient.swift
+++ b/FRCore/FRCore/Network/RestClient.swift
@@ -30,9 +30,11 @@ public class RestClient: NSObject {
     static var defaultURLSessionConfiguration: URLSessionConfiguration {
         get {
             let config = URLSessionConfiguration.default
+            // Setting `httpCookieStorage` to nil disables cookie storage
             config.httpCookieStorage = nil
             config.httpCookieAcceptPolicy = .never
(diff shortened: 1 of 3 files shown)


FRCore/FRCore/Network/RestClient.swift
```diff
@@ -33,6 +33,7 @@ public class RestClient: NSObject {
             config.httpCookieStorage = nil
             config.httpCookieAcceptPolicy = .never
             config.httpShouldSetCookies = false
+            config.urlCache = nil
```
george-bafaloukas-forgerock, line 36: Maybe let's add some comments here on why we are doing this, same for the cookieStorage

george-bafaloukas-forgerock (approved):

FRCore/FRCore/Network/RestClient.swift
vahancouver: Added comments

rodrigoareis (approved):
Changes looks good to me

spetrov (approved):
LGTM! 👍🏻

PR #243, opened 2023-10-24
Fix the issue in TelephonyCollector for dual sim phones iOS15 and less

vahancouver:
# JIRA Ticket

[SDKS-2776](https://bugster.forgerock.org/jira/browse/SDKS-2776)

# Description

Fix the issue in TelephonyCollector for dual sim phones iOS15 and less

### DIFF
--- a/FRAuth/FRAuth/Device/TelephonyCollector.swift
+++ b/FRAuth/FRAuth/Device/TelephonyCollector.swift
@@ -2,7 +2,7 @@
 //  TelephonyCollector.swift
 //  FRAuth
 //
-//  Copyright (c) 2019-2021 ForgeRock. All rights reserved.
+//  Copyright (c) 2019-2023 ForgeRock. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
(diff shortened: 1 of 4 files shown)


FRAuth/FRAuth/Device/TelephonyCollector.swift
```diff
@@ -28,35 +29,53 @@ public class TelephonyCollector: DeviceCollector {
...
+        var carrier: CarrierInfo?
         
-        if #available(iOS 12.0, *) {
-            if let providers = networkInfo.serviceSubscriberCellularProviders, providers.keys.count > 0 {
-                for (_, thisCarrier) in providers{
-                    carrier = thisCarrier
-                    break
-                }
+        
+        if let providers = networkInfo.serviceSubscriberCellularProviders, providers.keys.count > 0 {
```
jeyanthanperiyasamy, line 35: i think we can simplify this to one line may be ?

 var carriers = providers.map { (carrierName: $0.value.carrierName , isoCountryCode: $0.value.isoCountryCode ) }

jeyanthanperiyasamy (commented):
overall looks good, just a minor comment

jeyanthanperiyasamy (approved):
we need to have a ticket to either stop supporting this in future versions to avoid inconsistency between ios versions

spetrov (approved):
LGTM! 👍🏻

PR #247, opened 2023-11-23
Add custom claims to device binding and signing

vahancouver:
# JIRA Ticket

[SDKS-2788](https://bugster.forgerock.org/jira/browse/SDKS-2788)

# Description

[iOS] Add custom claim to signed Device Binding and Device Signing JWT


FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
```diff
@@ -81,9 +88,14 @@ extension DeviceAuthenticator {
     /// - Parameter userId: user Id received from server
     /// - Parameter challenge: challenge received from server
     /// - Parameter expiration: experation Date of jws
+    /// - Parameter customClaims: A dictionary of custom claims to be added to the jws payload
     /// - Returns: compact serialized jws
     /// - Throws: `DeviceBindingStatus` if any error occurs while signing
-    public func sign(keyPair: KeyPair, kid: String, userId: String, challenge: String, expiration: Date) throws -> String {
+    public func sign(keyPair: KeyPair, kid: String, userId: String, challenge: String, expiration: Date, customClaims: [String: Any] = [:]) throws -> String {
+        guard customClaims.isEmpty || validateCustomClaims(customClaims) else {
```
witrisna, line 95: - Move the check earlier, under DeviceBindingCallback and DeviceSigningVerifierCallback
- Instead of check empty first, can we move the check empty inside validateCustomClaims? if the list is empty consider valid.
- Since the caller is provide invalid argument, Shoud we throw invalidargumentexception?

FRDeviceBinding/FRDeviceBindingTests/DeviceAuthenticatorTests.swift
```diff
@@ -455,4 +455,166 @@ class DeviceAuthenticatorTests: FRBaseTestCase {
             //all good, do nothing
         }
     }
+    
+    
+    func test_14_ValidateCustomClaims_valid() {
```
witrisna, line 460: Add a test case that overrides the validateCustomClaims method, the validateCustomClaims method just return true, and when doing the binding or signing, the developer wants to override one of the existing claim.

witrisna (changes requested):

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
vahancouver: @witrisna 

- Moved _validateCustomClaims_ call into  **DeviceBindingCallback** and **DeviceSigningVerifierCallback**
- Removed the empty check all together, _validateCustomClaims_ method returns true for empty claims already
- In swift we don' have compile time invalidargumentexception the way java has, and also exceptions are handled differently in swift, they are runtime. I could call _fatalError()_ which is compile time error, but not sure that's what we want.

FRDeviceBinding/FRDeviceBindingTests/DeviceAuthenticatorTests.swift
vahancouver: @witrisna Added two custom classes, one always return true for validateCustomClaims and one always return false.
Added unit tests for those

rodrigoareis (approved):
Changes looks good to me

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
```diff
@@ -191,6 +203,22 @@ extension DeviceAuthenticator {
...
+        let registeredKeys = [DBConstants.sub,
+                              DBConstants.challenge,
+                              DBConstants.exp,
+                              DBConstants.iat,
+                              DBConstants.nbf,
+                              DBConstants.platform,
+                              DBConstants.iss]
+        let customKeys: [String] = Array(customClaims.keys)
+        let conflictingKeys = customKeys.filter { registeredKeys.contains($0) }
+        return conflictingKeys.isEmpty
```
jeyanthanperiyasamy, line 220: customClaims.keys.filter { registeredKeys.contains($0) }.isEmpty ?

vahancouver: done

jeyanthanperiyasamy (approved):

witrisna (approved):

rodrigoareis (approved):
Changes looks good to me

spetrov (approved):
Looks good to me! 👍🏻

PR #252, opened 2023-12-20
Sdks 2886 Fix Device Binding API access level

vahancouver:
# JIRA Ticket

[SDKS-2886](https://bugster.forgerock.org/jira/browse/SDKS-2886)

Fix Device Binding API access level

Fix device binding inheritance issues  

# Description

1. Make Binding protocol and methods public
2. Make all **DeviceAuthenticator** subclasses init methods public
3. Add Playground.swift in the example app for validating API Access level during compilation
4. Fix typos in ViewController file.
5.  Update CHANGELOG
6. Refactor Device Binding to fix the inheritance issue caused by Swift protocol's default implementation limitations

------------
Apart from the bug that we had, where our Authenticators had internal init methods, we also found another issue while testing this fix. 
Turns out Swift protocol's default implementation has an UNEXPECTED limitation during inheritance. 

## Here is a the Issue Description:

Protocol **A** defines function **x**.
**A** has implementation of **x** in the _protocol extension_.
Class **B** conforms **A**, but _doesn't provide_ implementation for **x**.
Subclass **C** inherits from **B**, and _provides implementation_ for **x**.
Instance of **C** typed as **A** calls **x**. 

Instead of the calling the **x** defined in **C**, Swift calls **x** defined in the _protocol extension_.
```
// Defined protocol.
protocol A {
    func x() -> Int
}
extension A {
    func x() -> Int {
        return 0
    }
}

// B conforms A but doesn't provide implementation for x
class B: A {}

class C: B {
    func x() -> Int {
        return 1
    }
}

// Failure cases.
B().a() // 0
C().a() // 1
(C() as A).a() // 0 # We would expect this to be 1. 

// D conforms A and provides implementation for x
class D: A {
    func x() -> Int {
        return 1
    }
}

class E: D {
    override func x() -> Int {
        return 2
    }
}

// Success cases.
D().x() // 1
(D() as A).x() // 1
E().x() // 2
(E() as A).x() // 2
```
Here is a [blog post](https://alejandromp.com/blog/protocol-extensions-and-subclasses/) explaining it and here is the [bug](https://github.com/apple/swift/issues/42725) that is still Open on swift's  github.

The way we fixed the issue was getting rid of protocol methods default implementations in protocol extensions. 
In case of **DeviceAuthenticator** we created a **DefaultDeviceAuthenticator** _class_ that provides the default implementation for protocol methods.
In case of **Binding** we removed the default method implementations in the extension and moved them into **DeviceBindingCallback** and **DeviceSigningVerifierCallback** classes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,13 +1,18 @@
-# Unreleased
+## [4.3.0]
 #### Added
-- Update Jailbreak Detectors [SDKS-2796]
-
-# Version 4.1.1
-
-## [4.1.1]
+- AppIntegrity Callback support [SDKS-2630] [SDKS-2761]
+- New ephemeralAuthSession browser type (iOS13+) [SDKS-2707]
+- `iat` and `nbf` claims to Device Binding jws payload [SDKS-2748]
+- Custom claims to device signing verifier [SDKS-2788]
 
 #### Fixed
 - Fixed an issue where the `issuer` parameter was not properly parsed when using AM 7.2.x [SDKS-2653]
+- Updated Jailbreak Detectors [SDKS-2796]
+- Fixed an issue related to Inadequate Cache Control [SDKS-2700]
+- Fixed the issue when sfViewController setting in Centralized login had the "entersReaderIfAvailable" as true [SDKS-2746] 
+- Fix the issue with DeviceProfile Collector affecting phones with dual sim cards in iOS 16.3 and earlier [SDKS-2776]
+- Improve unit and e2e tests [SDKS-2637]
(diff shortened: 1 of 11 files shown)


rodrigoareis (approved):
Changes looks good to me

spetrov (approved):
LGTM! 👍🏻

george-bafaloukas-forgerock (approved):
LGTM!

spetrov (approved):

rodrigoareis (approved):
Changed looks good to me

witrisna (approved):

PR #254, opened 2023-12-27
Fix the bug when device binding local key was not deleted

vahancouver:
# JIRA Ticket

[SDKS-2887](https://bugster.forgerock.org/jira/browse/SDKS-2887)

# Description

[iOS] Make sure iOS removes the device binding key record from local device repo on failure


rodrigoareis (approved):
Changes looks good to me

spetrov (approved):
Changes look good to me! 👍🏻

PR #282, opened 2024-06-03
FRURLProtocol ignores ssl pinning

george-bafaloukas-forgerock:
# JIRA Ticket

Please, link jira ticket here.](https://bugster.forgerock.org/jira/browse/SDKS-3239)

# Description

Fix for SSL Pinning and FRURLProtocol uses

### DIFF
--- a/FRAuth/FRAuth/Network/FRURLProtocol.swift
+++ b/FRAuth/FRAuth/Network/FRURLProtocol.swift
@@ -1,8 +1,8 @@
-// 
+//
 //  FRURLProtocol2.swift
 //  FRAuth
 //
-//  Copyright (c) 2020 ForgeRock. All rights reserved.
+//  Copyright (c) 2020-2024 ForgeRock. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
@@ -12,7 +12,7 @@
 import Foundation
 import FRCore
 
-@objc public class FRURLProtocol: URLProtocol {
+@objc open class FRURLProtocol: URLProtocol {
     
     //  MARK: - Public Property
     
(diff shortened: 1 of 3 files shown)


jeyanthanperiyasamy:
can we add some tests around this for both success and failure cases?

jeyanthanperiyasamy (changes requested):
comments provided

george-bafaloukas-forgerock:
> can we add some tests around this for both success and failure cases?

Might a bit hard to get proper success for SSL Pinning, unless we set on the tests the keyhashes of the test server

FRAuth/FRAuth/Network/FRURLProtocol.swift
```diff
@@ -20,6 +21,8 @@ import Foundation
     /// AuthorizationPolicy for URLProtocol
     @objc public static var authorizationPolicy: AuthorizationPolicy?
     
+    @objc public static var frSecurityConfiguration: FRSecurityConfiguration?
```
rodrigoareis, line 24: Missing doc

rodrigoareis (changes requested):
Overall changes looks good. Just minor comment to address

george-bafaloukas-forgerock:
> Overall looks good.
> 
> Can you please explain why we need this part (I see you added it later) else { completionHandler(.performDefaultHandling, nil) return } we didn't have this before adding the ssl pinning part.

Yes I had to add this, as I have added the 
```func urlSession(_ session: URLSession, task: URLSessionTask, didReceive challenge: URLAuthenticationChallenge, completionHandler: @escaping (URLSession.AuthChallengeDisposition, URLCredential?) -> Void)```
Method and on some requests it might return other types of `authenticationMethod`. We need to allow `defaultHandling` for the rest otherwise this will not call the `completion` and will be an API misuse. This was found through the test

vahancouver (approved):
LGTM

rodrigoareis (approved):
Looks good to me

PR #286, opened 2024-07-03
Fix the issue when session would get revoked twice during browser logout

vahancouver:
[SDKS-3301](https://bugster.forgerock.org/jira/browse/SDKS-3301)

[iOS] FRUser.logout(...) should not invoke endSession when oauthSignoutRedirectUri is present in the configuration


spetrov (approved):
LGTM! 👍🏻

PR #314, opened 2024-12-12
SDKS-3533 Prevent duplicated notification on iOS SDK

rodrigoareis:
# JIRA Ticket

[SDKS-3533](https://pingidentity.atlassian.net/browse/SDKS-3533) [iOS][Authenticator] Prevent duplicated notification on iOS SDK

# Description

The iOS SDK does not verify whether a push notification with an identical identifier has already been stored, delegating this responsibility to the application developer. This PR include the following changes:
- Added a new `getNotificationByMessageId` method to the `StorageClient` and `FRAClient` 
- Updated the `FRAPushHandler` class to check whether a payload was already persisted
- Update unit tests

# Definition of Done Checklist:

- [ ] Acceptance criteria is met.
- [ ] All tasks listed in the user story have been completed.
- [ ] Coded to standards.
- [ ] Ensure backward compatibility.
- [ ] API reference docs is updated.
- [ ] Unit tests are written.
- [ ] Integration tests are written.
- [ ] e2e tests are written.
- [ ] Functional spec is written/updated.
- [ ] Example code snippets have been added.
- [ ] Change log updated.
- [ ] Documentation story is created and tracked.
- [ ] Tech debts and remaining tasks are tracked in separated ticket(s).

### DIFF
--- a/FRAuthenticator/FRAuthenticator/Push/FRAPushHandler.swift
+++ b/FRAuthenticator/FRAuthenticator/Push/FRAPushHandler.swift
@@ -99,6 +99,7 @@ public class FRAPushHandler: NSObject {
                 }
                 else {
                     FRALog.w("PushNotification object failed to be stored into StorageClient")
+                    return nil
                 }
                 
                 return notification
--- a/FRAuthenticator/FRAuthenticator/Push/FRAPushHandler.swift
+++ b/FRAuthenticator/FRAuthenticator/Push/FRAPushHandler.swift
@@ -2,7 +2,7 @@
 //  FRAPushHandler.swift
 //  FRAuthenticator
 //
-//  Copyright (c) 2020 ForgeRock. All rights reserved.
+//  Copyright (c) 2020-2024 Ping Identity. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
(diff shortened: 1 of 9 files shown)


FRAuthenticator/FRAuthenticator/Push/FRAPushHandler.swift
```diff
@@ -68,35 +68,41 @@ public class FRAPushHandler: NSObject {
...
-                FRALog.v("PushNotification object is created and saved into StorageClient")
-            }
-            else {
-                FRALog.w("PushNotification object failed to be stored into StorageClient")
+                
+                if FRAClient.storage.setNotification(notification: notification) {
+                    FRALog.v("PushNotification object is created and saved into StorageClient")
+                }
+                else {
+                    FRALog.w("PushNotification object failed to be stored into StorageClient")
```
george-bafaloukas-forgerock, line 101: Should we not treat that as an error and return nil? If we end up on this situation, is the Push actionable or would the app be stuck?

rodrigoareis: good catch! we should return nil here. It's unlikely we will have an exception while persisting it because the default storage is the Keychain. However, if a developer decide to use SQLite or any other storage, it may happen, so we need to confirm

george-bafaloukas-forgerock (approved):
LGTM

vahancouver (approved):
Changes look good to me

jeyanthanperiyasamy (approved):

PR #320, opened 2025-01-20
Midships 4.7.0

yuxiang-midships:
Hi Ping SDK Team,

Yuxiang here from Midships. I am raising this pull request to show some changes we are doing to support some of our clients for Ping SDK integration. These change on iOS SDK is to allow certain feature available on the Android SDK to be also available on the iOS SDK.

Summary:
1. expose the token manager interface and instance on iOS sdk
2. expose the clear method in token manager on the iOS sdk
3. add the persist token method in token manager on the iOS sdk 

Feel free to reach out to me for more details and why some clients will need these changes.

Best regards
Yuxiang
Associate Director, Midships

### DIFF
--- a/FRAuth/FRAuth/FRAuth.swift
+++ b/FRAuth/FRAuth/FRAuth.swift
@@ -50,7 +50,7 @@ public final class FRAuth: NSObject {
     /// OAuth2Client instance for FRAuth; OAuth2Client values are retrieved from .plist configuration file
     var oAuth2Client: OAuth2Client?
     /// TokenManager instance for FRAuth to perform any token related operation
-    var tokenManager: TokenManager?
+    public var tokenManager: TokenManager?
     /// SessionManager instance for FRAuth to perform manage, and persist session
     var sessionManager: SessionManager
     /// KeychainManager instance for FRAuth to perform any keychain related operation to persist and/or retrieve credentials
--- a/FRAuth/FRAuth/Manager/TokenManager.swift
+++ b/FRAuth/FRAuth/Manager/TokenManager.swift
@@ -12,7 +12,7 @@ import Foundation
 
 
 /// TokenManager class is a management class responsible for persisting, retrieving, and refreshing OAuth2 token(s)
-struct TokenManager {
+public struct TokenManager {
     
     var oAuth2Client: OAuth2Client
     var keychainManager: KeychainManager
@@ -22,7 +22,7 @@ struct TokenManager {
     /// - Parameters:
     ///   - oAuth2Client: OAuth2Client instance for OAuth2 token protocols
     ///   - keychainManager: KeychainManager instance for secure credentials management
-    public init(oAuth2Client: OAuth2Client, keychainManager: KeychainManager) {
+    init(oAuth2Client: OAuth2Client, keychainManager: KeychainManager) {
         self.oAuth2Client = oAuth2Client
         self.keychainManager = keychainManager
     }
--- a/FRAuth/FRAuth/Manager/TokenManager.swift
+++ b/FRAuth/FRAuth/Manager/TokenManager.swift
@@ -357,7 +357,7 @@ public struct TokenManager {
     
     
     /// Clears all credentials locally as there is no more valid credentials to renew user's session
-    func clearCredentials() {
+    public func clearCredentials() {
         self.keychainManager.cookieStore.deleteAll()
         let _ = try? self.keychainManager.setAccessToken(token: nil)
         self.keychainManager.setSSOToken(ssoToken: nil)
--- a/FRAuth/FRAuth/Manager/TokenManager.swift
+++ b/FRAuth/FRAuth/Manager/TokenManager.swift
@@ -14,7 +14,7 @@ import Foundation
 /// TokenManager class is a management class responsible for persisting, retrieving, and refreshing OAuth2 token(s)
 public struct TokenManager {
     
-    var oAuth2Client: OAuth2Client
+    public var oAuth2Client: OAuth2Client
     var keychainManager: KeychainManager
     
     /// Initializes TokenMAnager instance with optional Access Group for shared Keychain Group identifier
--- a/FRAuth/FRAuth/Manager/TokenManager.swift
+++ b/FRAuth/FRAuth/Manager/TokenManager.swift
@@ -366,6 +366,10 @@ public struct TokenManager {
         Browser.currentBrowser = nil
     }
     
+    public func persistToken(token: AccessToken) {
+        let _ = try? self.keychainManager.setAccessToken(token: token)
+    }
+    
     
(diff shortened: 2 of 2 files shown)


george-bafaloukas-forgerock:
Hi @yuxiang-midships thanks for raising this PR. 

Can you please provide more information on the use cases you want to cover by making the Token manager and the "persist token method" public?

yuxiang-midships:
https://github.com/ForgeRock/forgerock-ios-sdk/pull/320#issuecomment-2651054630

Hi George,

Thank you for taking a look at this. We are using FRSession instead of FRuser to have more control over different authentication tree through out the mobile apple CIAM journeys. At the same time, we have Oauth2 flows that will need to be supported. For the android SDK, we can get around and achieve the logic we want as we can access:

Config.getInstance().getSessionManager().getTokenManager()

which allow us to access on android the tokenManager.

We can then use the exchangeToken, clear and persist method in the token manager class to achieve using FR session for invoking AM trees and then use the tokenID to do oauth2 flows.

However when it come to the iOS SDK, we are not able to access the token manager and use the equivalent method. Therefore we modified the iOS SDK as illustrated in the commits to allow iOS sdk to have the same feature as the android SDK

We make the token manager and the ouath2 client in the token manager public. and we also make the clearCredential method in token manager public. Lastly we added the persist method to the token manager.

george-bafaloukas-forgerock:
Hi there,

Thanks for your reply. I will be actively reviewing this on the current Sprint. I am assuming this is related with the support case: `01076030` as well. Please expect more questions as part of that :)

george-bafaloukas-forgerock:
Hey Yuxiang and team, one more clarification. 
You mention
> We are using FRSession instead of FRuser to have more control over different authentication tree through out the mobile apple CIAM journeys

This is correct through the use of FRSession.authenticate you can control the use of the `authentication tree`. At the end of the authentication you can then jump on the FRUser classes. For example check out the following code:

`func handleNode(token: Token?, node: Node?, error: Error?) {
        if (token == nil) && (node == nil) && (error == nil) {
            print("All nil indicates a noSession=true completed Journey")
            self.updateStatus()
            return
        }
        if error != nil {
            print(error!)
            return
        }
        self.currentNode = node
        if let _ = token {
            print("Session only Journey is now complete")
            FRUser.currentUser?.getAccessToken(completion: { user, error in
                DispatchQueue.main.async {
                    self.updateStatus()
                }
            })
        } else {
            if node == nil {
                return
            }
            for callback: Callback in node!.callbacks {
            }
        }
    }`
So basically after getting the `Token` you can work with the FRUser interface which should expose all the methods required. Please let me know if I am missing something here. The only possible reason I can think of exposing more from the TokenManager would be the use of different OAuth2.0 flows other than the one the SDK supports OOTB. 

Same logic applies to Android

yuxiang-midships:
Hi George,

Thanks for reaching out. You are absolutely correct that after using the FRSession to finish the tree and get a SSO token, when can then use FR user to exchange the SSO token for Ouath2 Access token. However there are 2 issues we faced.

yuxiang-midships:
Hi George,

You are absolutely correct that after using FRSession, we can use FRUser to exchange for Oauth2 Access token. However two of our clients are also using AM trees to carryout extra authentication at a transaction level (not session level). In the flow we implemented, we have custom Ouath2 scripts that enable issuance of special scoped access token where the mobile end can use to invoke certain high risk transactions.

FRAuth/FRAuth/Manager/TokenManager.swift
```diff
@@ -12,17 +12,17 @@ import Foundation
 
 
 /// TokenManager class is a management class responsible for persisting, retrieving, and refreshing OAuth2 token(s)
-struct TokenManager {
+public struct TokenManager {
     
-    var oAuth2Client: OAuth2Client
+    public var oAuth2Client: OAuth2Client
```
george-bafaloukas-forgerock, line 17: Add code doc for the public oAuth2Client

FRAuth/FRAuth/Manager/TokenManager.swift
```diff
@@ -366,6 +366,10 @@ struct TokenManager {
         Browser.currentBrowser = nil
     }
     
+    public func persistToken(token: AccessToken) {
```
george-bafaloukas-forgerock, line 369: add code doc for the persistToken method

george-bafaloukas-forgerock (changes requested):
Hello @yuxiang-midships  and team, thank you very much for submitting this request and providing these clarifications. 

Can I please ask you to have a look on the comments and also make sure that any follow up commits are done using "Verified" account status. Please let us know if you need details on how to set up a "Verified" key.

Last but not least, can you please add some tests for the proposed changes?

Kind regards,
George Bafaloukas

FRAuth/FRAuth/Manager/TokenManager.swift
```diff
@@ -12,17 +12,17 @@ import Foundation
 
 
 /// TokenManager class is a management class responsible for persisting, retrieving, and refreshing OAuth2 token(s)
-struct TokenManager {
+public struct TokenManager {
```
vahancouver, line 15: Please update the copyright header to reflect the latest changes year

FRAuth/FRAuth/FRAuth.swift
```diff
@@ -50,7 +50,7 @@ public final class FRAuth: NSObject {
     /// OAuth2Client instance for FRAuth; OAuth2Client values are retrieved from .plist configuration file
     var oAuth2Client: OAuth2Client?
     /// TokenManager instance for FRAuth to perform any token related operation
-    var tokenManager: TokenManager?
+    public var tokenManager: TokenManager?
```
vahancouver, line 53: Please update the copyright header to reflect the latest changes year

FRAuth/FRAuth/Manager/TokenManager.swift
```diff
@@ -12,17 +12,17 @@ import Foundation
...
+    public var oAuth2Client: OAuth2Client
     var keychainManager: KeychainManager
     
     /// Initializes TokenMAnager instance with optional Access Group for shared Keychain Group identifier
     ///
     /// - Parameters:
     ///   - oAuth2Client: OAuth2Client instance for OAuth2 token protocols
     ///   - keychainManager: KeychainManager instance for secure credentials management
-    public init(oAuth2Client: OAuth2Client, keychainManager: KeychainManager) {
+    init(oAuth2Client: OAuth2Client, keychainManager: KeychainManager) {
```
vahancouver, line 25: was public meant to be removed here?

vahancouver (changes requested):
Left some comments to address

PS: the commits need to be signed

vahancouver:
stalled with unaddressed change requests since Feb 2025

PR #341, opened 2025-07-09
feat: Implement ResponseInterceptor

sakinsete:
# Description
This commit introduces the `ResponseInterceptor` protocol and its registry, enabling developers to intercept and modify SDK-generated responses.

Key changes:
- Added `ResponseInterceptor.swift` defining the protocol for response interception.
- Added `ResponseInterceptorRegistry.swift` in FRCore to manage an array of `ResponseInterceptor` instances.
- Added `FRResponseInterceptorRegistry.swift` in FRAuth as a wrapper for the FRCore registry.
- Updated `RestClient.swift` to incorporate response interception logic, similar to existing request interception.
- Made `Response` struct properties public for accessibility.
- Renamed `interceptors` to `requestInterceptors` in `RestClient.swift` for clarity.
- Updated project files to include the new response interceptor files.


george-bafaloukas-forgerock:
Hi @sakinsete thank you for creating this PR. We will be reviewing and posting our updates here. Apologies for the late reply. Can you please also provide some further details and a use case on why this feature is needed and the use cases you are trying to solve?

Kind regards,
George

vahancouver:
Hi @sakinsete — thank you for this contribution and for your patience while we worked through the review.

  The forgerock-ios-sdk (this repo) is in maintenance mode: we're keeping it healthy with bug and security fixes, but we're no longer accepting new public API surface here. Feature investment has moved to the new Ping iOS SDK (https://github.com/ForgeRock/ping-ios-sdk), which is where we'd like to direct work like this going forward.

  The good news is that your use case — observing SDK-issued HTTP responses to feed them into another system (F5 Shape, telemetry, logging, etc.) — is already supported natively in the new SDK. The Network module exposes a response interceptor closure on HttpClientConfig:

```
  let client = HttpClient.createClient { config in
      config.onResponse { response in
          // response.status, response.body, response.getHeader(...), etc.
          // hand off to Shape SDK / telemetry / logging
      }
  }
```

 Given that the new SDK already covers this, we're going to close this PR. Really appreciate you taking the time to put it together and to follow up multiple times — sorry we didn't get you a clearer answer sooner. If you hit anything missing in the new SDK's interception model, please open an issue there and we'll take a look.

  Thanks again,
  Vahan

PR #344, opened 2025-08-08
Rename Data extension bytes property to bytesArray to avoid Xcode26 issues

vahancouver:
[SDKS-4299](https://pingidentity.atlassian.net/browse/SDKS-4299) [SPIKE] [iOS] Validate Legacy and Unified SDKs for Xcode 26

Rename Data extension bytes property to bytesArray to avoid Xcode26 issues


george-bafaloukas-forgerock (approved):
lgtm

PR #345, opened 2025-08-25
Fix UserInfo and Address decoding

vahancouver:
[SDKS-4346](https://pingidentity.atlassian.net/browse/SDKS-4346) [iOS] [Legacy SDK] [CWE-502] Fix Decoding Methods in NSSecureCoding Implementation

1. Fix decoding for Address and UserInfo
2. Add Unit tests

PS. During implementation I noticed the following:
1. Archiving/Unarchiving is never used for those classes in the SDK
3. Address class had a bug, and would always initialize to nil from a decoder (but as it isn't used we never caught it)


george-bafaloukas-forgerock (approved):
LGTM

spetrov (approved):
LGTM! 👍🏻

PR #362, opened 2026-02-23
fix: preserve credentials on timeout and offline refresh errors

naveengaur800:
This PR fixes an auth reliability issue where temporary network failures during token renewal could force a full logout.

## Problem

 When getAccessToken() triggered renewal and the request failed due to transport conditions (no internet or timeout), the flow treated that failure like an unrecoverable auth rejection and cleared local auth state.

  That converted a transient outage into a permanent logout, requiring full re-authentication even
  though credentials could still be valid.

  ## Context

  - Clearing tokens on transport failure degrades user experience and causes unnecessary re-login.
  - SDK consumers need the original failure reason to handle retry UX correctly.

  ## What changed

For transport-level failures (offline/timeout), token renewal is now non-destructive:
  - local auth state is preserved
  - the original network error is returned to the caller

Behavior for true authentication failures (for example invalid/unauthorized token scenarios) is unchanged.

  ## Impact

  Users are no longer logged out because of temporary network problems, and clients can show retry UX based on the actual failure.


spetrov:
These issues have been addressed by #387 and the fixes will be available with the next release of the SDK.

PR #370, opened 2026-04-10
SDKS-4922 - Improving the Biometric_allow_fallback to respect the PIN if the biom…

george-bafaloukas-forgerock:
Improving the Biometric_allow_fallback to respect the PIN if the biometrics are not set on a device

# JIRA Ticket

SDKS-4922

# Description

## Core Changes
State Tracking: The app now captures the biometricDomainState (a unique hash of the biometric database) during the initial device binding and stores it within the UserKey object.

Validation on Sign-In: Every time a signing operation occurs, the current biometric state is compared against the stored state.

Security Enforcement: If the states do not match—indicating the biometric set has changed—the app automatically deletes the existing keys and throws a .clientNotRegistered error, forcing the user to re-bind for security.

Access Control Update: The security flags were changed from .biometryCurrentSet to .biometryAny. This moves the responsibility of invalidating keys from the iOS system to the application, allowing for a more controlled user experience (e.g., custom error messages or specific recovery flows).

# Definition of Done Checklist:

- [X] Coded to standards.
- [X] Ensure backward compatibility.
- [X] API reference docs is created or updated, if applicable.
- [X] Unit tests are written or updated.
- [X] Integration tests are written, if applicable.

### DIFF
--- a/FRDeviceBinding/FRDeviceBinding/Sources/Callbacks/DeviceBindingCallback.swift
+++ b/FRDeviceBinding/FRDeviceBinding/Sources/Callbacks/DeviceBindingCallback.swift
@@ -2,7 +2,7 @@
 //  DeviceBindingCallback.swift
 //  FRDeviceBinding
 //
-//  Copyright (c) 2022 - 2025 Ping Identity Corporation. All rights reserved.
+//  Copyright (c) 2022 - 2026 Ping Identity Corporation. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
--- a/FRDeviceBinding/FRDeviceBinding/Sources/Callbacks/DeviceBindingCallback.swift
+++ b/FRDeviceBinding/FRDeviceBinding/Sources/Callbacks/DeviceBindingCallback.swift
@@ -193,7 +193,7 @@ open class DeviceBindingCallback: MultipleValuesCallback, Binding {
         
         do {
             let keyPair = try authInterface.generateKeys()
-            userKey = UserKey(id: keyPair.keyAlias, userId: userId, userName: userName, kid: UUID().uuidString, authType: deviceBindingAuthenticationType, createdAt: Date().timeIntervalSince1970)
+            userKey = UserKey(id: keyPair.keyAlias, userId: userId, userName: userName, kid: UUID().uuidString, authType: deviceBindingAuthenticationType, createdAt: Date().timeIntervalSince1970, biometricDomainState: authInterface.biometricDomainState())
             guard let userKey = userKey else {
                 throw DeviceBindingStatus.unsupported(errorMessage: "Cannot create userKey")
             }
--- a/FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
+++ b/FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
@@ -53,6 +53,9 @@ public protocol DeviceAuthenticator {
     /// Get the Device Binding Authentication Type
     func type() -> DeviceBindingAuthenticationType
     
+    /// Get the current biometric domain state for detecting enrollment changes
+    func biometricDomainState() -> Data?
+    
     /// initialize already created entity with useriD and Promp
     /// - Parameter userId: userId of the authentication
     /// - Parameter prompt: Prompt containing the description for authentication
@@ -100,6 +103,11 @@ open class DefaultDeviceAuthenticator: DeviceAuthenticator {
     /// Get the Device Binding Authentication Type
     open func type() -> DeviceBindingAuthenticationType {
         return .none
+    }
+    
+    /// Get the current biometric domain state. Returns nil by default.
+    open func biometricDomainState() -> Data? {
+        return nil
     }
     
     /// Remove Keys
@@ -370,15 +378,39 @@ open class BiometricAndDeviceCredential: BiometricAuthenticator {
     /// Access Control for the authetication type
     open override func accessControl() -> SecAccessControl? {
 #if !targetEnvironment(simulator)
-        return SecAccessControlCreateWithFlags(kCFAllocatorDefault, kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly, [.biometryCurrentSet, .or, .devicePasscode, .privateKeyUsage], nil)
+        return SecAccessControlCreateWithFlags(kCFAllocatorDefault, kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly, [.biometryAny, .or, .devicePasscode, .privateKeyUsage], nil)
 #else
-        return SecAccessControlCreateWithFlags(kCFAllocatorDefault, kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly, [.biometryCurrentSet, .or, .devicePasscode], nil)
+        return SecAccessControlCreateWithFlags(kCFAllocatorDefault, kSecAttrAccessibleWhenPasscodeSetThisDeviceOnly, [.biometryAny, .or, .devicePasscode], nil)
 #endif
     }
     
     
     open override func type() -> DeviceBindingAuthenticationType {
         return .biometricAllowFallback
+    }
+    
+    
+    /// Returns the current biometric domain state from LAContext.
+    /// Used to detect biometric enrollment changes between bind and sign operations.
+    open override func biometricDomainState() -> Data? {
+        let context = LAContext()
+        context.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: nil)
+        return context.evaluatedPolicyDomainState
+    }
+    
+    
+    /// Override sign to validate biometric enrollment has not changed since binding.
+    /// If biometrics were enrolled at bind time and the enrollment has since changed,
+    /// the keys are deleted and `.clientNotRegistered` is thrown to trigger re-binding.
+    open override func sign(userKey: UserKey, challenge: String, expiration: Date, customClaims: [String: Any] = [:]) throws -> String {
+        if let storedState = userKey.biometricDomainState {
+            let currentState = biometricDomainState()
+            if currentState != storedState {
+                deleteKeys()
+                throw DeviceBindingStatus.clientNotRegistered
+            }
+        }
+        return try super.sign(userKey: userKey, challenge: challenge, expiration: expiration, customClaims: customClaims)
     }
 }
 
--- a/FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
+++ b/FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
@@ -340,7 +340,26 @@ open class BiometricOnly: BiometricAuthenticator {
 }
 
 
-/// DeviceAuthenticator adoption for biometric and Device Credential authentication
+/// DeviceAuthenticator adoption for biometric and Device Credential authentication.
+///
+/// This authenticator supports the `BIOMETRIC_ALLOW_FALLBACK` policy, allowing authentication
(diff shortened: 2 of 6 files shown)


witrisna (approved):

FRDeviceBinding/FRDeviceBinding/Sources/UserDeviceKeyService.swift
```diff
@@ -105,4 +105,6 @@ public struct UserKey: Equatable, Codable {
     public var kid: String
     public var authType: DeviceBindingAuthenticationType
     public var createdAt: Double
+    /// The biometric domain state captured at bind time, used to detect biometric enrollment changes
```
rodrigoareis, line 114: Update copyright

FRDeviceBinding/FRDeviceBinding/Sources/Callbacks/DeviceBindingCallback.swift
```diff
@@ -193,7 +193,7 @@ open class DeviceBindingCallback: MultipleValuesCallback, Binding {
         
         do {
             let keyPair = try authInterface.generateKeys()
-            userKey = UserKey(id: keyPair.keyAlias, userId: userId, userName: userName, kid: UUID().uuidString, authType: deviceBindingAuthenticationType, createdAt: Date().timeIntervalSince1970)
+            userKey = UserKey(id: keyPair.keyAlias, userId: userId, userName: userName, kid: UUID().uuidString, authType: deviceBindingAuthenticationType, createdAt: Date().timeIntervalSince1970, biometricDomainState: authInterface.biometricDomainState())
```
rodrigoareis, line 196: Update copyright

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
```diff
@@ -53,6 +53,9 @@ public protocol DeviceAuthenticator {
     /// Get the Device Binding Authentication Type
     func type() -> DeviceBindingAuthenticationType
     
+    /// Get the current biometric domain state for detecting enrollment changes
```
rodrigoareis, line 56: Consider add new tests to cover the changes.

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
```diff
@@ -367,19 +380,55 @@ open class BiometricAndDeviceCredential: BiometricAuthenticator {
...
     open override func type() -> DeviceBindingAuthenticationType {
         return .biometricAllowFallback
     }
+    
+    
+    /// Returns the current biometric domain state from LAContext.
+    /// Used to detect biometric enrollment changes between bind and sign operations.
+    open override func biometricDomainState() -> Data? {
+        let context = LAContext()
+        context.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: nil)
```
rodrigoareis:
```suggestion
        _ = context.canEvaluatePolicy(.deviceOwnerAuthenticationWithBiometrics, error: nil)
```

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
```diff
@@ -53,6 +53,9 @@ public protocol DeviceAuthenticator {
     /// Get the Device Binding Authentication Type
     func type() -> DeviceBindingAuthenticationType
     
+    /// Get the current biometric domain state for detecting enrollment changes
+    func biometricDomainState() -> Data?
+    
     /// initialize already created entity with useriD and Promp
```
rodrigoareis:
```suggestion
    /// initialize already created entity with userID and Prompt
```

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
```diff
@@ -367,19 +380,55 @@ open class BiometricAndDeviceCredential: BiometricAuthenticator {
     }
     
     
-    /// Access Control for the authetication type
+    /// Access Control for the authetication type.
```
rodrigoareis:
```suggestion
    /// Access Control for the authentication type.
```

rodrigoareis: This typo `"authetication" -> "authentication"` is repeated in other locations in this file

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
```diff
@@ -53,6 +53,9 @@ public protocol DeviceAuthenticator {
     /// Get the Device Binding Authentication Type
     func type() -> DeviceBindingAuthenticationType
     
+    /// Get the current biometric domain state for detecting enrollment changes
+    func biometricDomainState() -> Data?
```
rodrigoareis, line 57: I believe any third-party or customer code that implements `DeviceAuthenticator` directly (not subclassing `DefaultDeviceAuthenticator`) will now fail to compile. While `DefaultDeviceAuthenticator` provides a default implementation, the protocol itself doesn't have a default extension.

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
```diff
@@ -367,19 +380,55 @@ open class BiometricAndDeviceCredential: BiometricAuthenticator {
...
     
-    /// Access Control for the authetication type
+    /// Access Control for the authetication type.
+    /// Dynamically selects flags based on biometric availability:
+    /// - When biometrics are enrolled: `.biometryAny OR .devicePasscode`
+    /// - When only passcode is set: `.devicePasscode` only
+    /// This avoids Secure Enclave rejecting key creation when biometry flags
+    /// are present but no biometrics are enrolled.
     open override func accessControl() -> SecAccessControl? {
+        let laContext = LAContext()
```
rodrigoareis, line 417: An `LAContext` is already created in ` generateKeys()`. Then accessControl() is also called at line 354, which creates another `LAContext`. This means two separate `LAContext` instances are created and `canEvaluatePolicy` is called twice for every key generation. Consider passing the biometric availability result or the `LAContext` to avoid double evaluation.

rodrigoareis (changes requested):
The fix solves the reported issue and the key creation no longer fails when biometrics are not enrolled but passcode is available. However, I still think that switching from `.biometryCurrentSet` to `.biometryAny` changes the security behavior:

- Before (.biometryCurrentSet): The Secure Enclave hardware invalidates the key automatically when biometric enrollment changes (e.g., a fingerprint is added/removed). This is a tamper-proof guarantee.

- After (.biometryAny): The key survives biometric enrollment changes at the hardware level. We are compensating with a software-level check using `evaluatedPolicyDomainState`, but this is weaker because:
  - It's enforced in SDK code, not by the Secure Enclave.
  - The sign override that performs the check is marked open, so subclasses can bypass it.
  - The check only covers the `sign(userKey:...)` path (used during signing verification), not the `sign(keyPair:...)` path (used during initial binding).

Bottom line: 
For `BIOMETRIC_ALLOW_FALLBACK` this trade-off is reasonable as the policy **already accepts passcode as a valid alternative**. **The Android SDK also doesn't enforce hardware-level biometric enrollment checks.** (@witrisna please confirm). But the change should be documented so consumers understand the shifted security model.

george-bafaloukas-forgerock:
@rodrigoareis Valid points. I will be adding some extra comments to document this and also we might want to note that in the docs. 
One clarification on the point below:
>   * The check only covers the `sign(userKey:...)` path (used during signing verification), not the `sign(keyPair:...)` path (used during initial binding).

"The check only covers the sign(userKey:...) path (used during signing verification), not the sign(keyPair:...) path (used during initial binding)."

This is not actually a gap. The sign(keyPair:...) path is only called during initial binding in DeviceBindingCallback.execute() — immediately after generateKeys() creates a fresh key pair. At that moment the biometricDomainState is captured and stored in the UserKey. There's no prior state to compare against because the key didn't exist yet. The domain state check is only meaningful on subsequent signing operations (the sign(userKey:...) path).

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
george-bafaloukas-forgerock: Fixed some further typos

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
george-bafaloukas-forgerock: fixed all instances

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
george-bafaloukas-forgerock: `accessControl()` is a public protocol method called independently from multiple places (tests, ApplicationPinDeviceAuthenticator, external consumers). It must be self-contained — it can't rely on receiving external state. Adding a parameter or internal state would either break the protocol contract or create coupling where the caller must remember to pre-evaluate biometrics.
In any case the benefit is negligible as `canEvaluatePolicy` and the `LAContext` allocation won't save anything on performance. Lastly the two contexts serve different purposes: one determines access control flags (security policy), the other determines whether to attach a pre-authentication context (UX behavior). Merging them would conflate two concerns.

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
george-bafaloukas-forgerock: Added an extension with default implementation. Good catch that one, but it should now compile fine after updating the SDK

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
george-bafaloukas-forgerock: Added more test cases to improve coverage. With binding we have to take into account as well that the tests run on a simulator that the enclave won't be accessible

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
```diff
@@ -367,19 +407,55 @@ open class BiometricAndDeviceCredential: BiometricAuthenticator {
...
+        return context.evaluatedPolicyDomainState
+    }
+    
+    
+    /// Override sign to validate biometric enrollment has not changed since binding.
+    /// If biometrics were enrolled at bind time and the enrollment has since changed,
+    /// the keys are deleted and `.clientNotRegistered` is thrown to trigger re-binding.
+    open override func sign(userKey: UserKey, challenge: String, expiration: Date, customClaims: [String: Any] = [:]) throws -> String {
+        if let storedState = userKey.biometricDomainState {
+            let currentState = biometricDomainState()
```
vahancouver, line 451: what happens if biometricDomainState() returns nil here?

vahancouver (approved):
Approved with one question.

FRDeviceBinding/FRDeviceBinding/Sources/DeviceBindingAuthenticators.swift
george-bafaloukas-forgerock: `DeviceBindingStatus.clientNotRegistered` will be thrown. In the implementation of this authenticator that will mean that the biometric has changed, which is correct given that we have a stored state. 
A re-binding will be required

rodrigoareis (approved):
Changes looks good to me

spetrov (approved):
LGTM! 👍🏻

PR #371, opened 2026-04-15
SDKS-3045 Updating SSO token mismatch logic & Step-up logic updates

george-bafaloukas-forgerock:
# JIRA Ticket

Please link jira ticket here, if applicable.

# Description

[...]

PR #374, opened 2026-04-23
feat: exclude revoke session call when new sso token received

harshvardhanarora:
Post the AM update that happened on the morning of 15th April 2026, few of the flows in the SDK broke down. It was later discovered that the cause behind it was the new session token that gets issued at the end of an authentication tree gets revoked because it is attached to the end session request that is made after a new token is received. This behaviour is a mis-match between the iOS and Android SDK and the Android SDK only makes the /revoke request and not the /endSession request whereas the iOS SDK makes them both. This change fixes that to align the behaviour between the 2 SDK and fix the issue caused after the backend update.

### DIFF
--- a/FRAuth/FRAuth/Authentication/Node.swift
+++ b/FRAuth/FRAuth/Authentication/Node.swift
@@ -293,7 +293,7 @@ public class Node: NSObject {
                             var revokeError: Error?
                             
                             // 2. Call the async function.
-                            tokenManager.revokeAndEndSession { error in
+                            tokenManager.revoke { error in
                                 FRLog.i("OAuth2 token set revocation finished. Error: \(error?.localizedDescription ?? "none")")
                                 // Capture the result and signal completion.
                                 revokeError = error
--- a/FRAuth/FRAuth/Manager/TokenManager.swift
+++ b/FRAuth/FRAuth/Manager/TokenManager.swift
@@ -191,6 +191,7 @@ struct TokenManager {
                 self.oAuth2Client.revoke(accessToken: token) { (error) in
                     completion(error)
                 }
+                completion(nil)
             }
             else {
                 completion(TokenError.nullToken)


george-bafaloukas-forgerock:
Hi there, thanks for your contribution. 

We are planning on addressing this issue as part of the https://github.com/ForgeRock/forgerock-ios-sdk/pull/371 change.
Can you please let us know what are the timelines you are looking for this to be included in a release and if you would be happy to entertain a BETA release for testing this out?

Kind regards,
George

harshvardhanarora:
> Hi there, thanks for your contribution.
> 
> We are planning on addressing this issue as part of the #371 change. Can you please let us know what are the timelines you are looking for this to be included in a release and if you would be happy to entertain a BETA release for testing this out?
> 
> Kind regards, George

Hi George

We are planned for go live next week on iOS and would need this change urgently to test it before we release. Could you please provide a beta version as soon as possible so we could test that on our end as well?

Regards,
Harshvardhan Arora

george-bafaloukas-forgerock:
> > Hi there, thanks for your contribution.
> > We are planning on addressing this issue as part of the #371 change. Can you please let us know what are the timelines you are looking for this to be included in a release and if you would be happy to entertain a BETA release for testing this out?
> > Kind regards, George
> 
> Hi George
> 
> We are planned for go live next week on iOS and would need this change urgently to test it before we release. Could you please provide a beta version as soon as possible so we could test that on our end as well?
> 
> Regards, Harshvardhan Arora

Hey @harshvardhanarora  as part of the support case raised we have provided the below steps. Please let us know if these work for you as a workaround. I would suggest we aim to keep the comms on this issue via the Ping Support case raised.
Summary of steps related to the issue:

- Call authenticate endpoint - to start auth journey
- Once journey is completed successfully - SessionTokenA gets stored in the app
- Call get access token using SessionTokenA - AccessTokenA, IdTokenA gets stored in the app 
- User is prompted to step upCall authenticate endpoint - start auth journey (this call now includes SessionTokenA as a cookie)
-- Once journey is completed successfully - SessionTokenB gets stored in the app
-- App terminated old session using IdTokenA
-- But because SessionA and SessionB are linked, both sessions are terminated

But, if you call Step 4.a without including the SessionTokenA cookie, in Step 4.c, only SessionA gets terminated.

There are 2 workarounds for this depending on what approach the customer wants to take up until we provide a fix in a feature version to the SDK.
Option one: Do not use the openid scope so that an IDToken is not used. This should work, as long as JLR is not using the IDToken in any part of the application. This would basically stop the unwanted revocation. 

 

Option two: Use a request interceptor that doesn’t send the first SSO Token to the step up Journey in the form of the Cookie. This will basically stop the link between the 2 SSO Tokens. As long as the first SSO Token is not required in the “step up” journey that solution will work as well. In order to do this in the iOS application the following example RequestInterceptor can be used:



```
// MARK: - Request Interceptors
class ForceAuthInterceptor: RequestInterceptor {
    func intercept(request: FRCore.Request, action: Action) -> FRCore.Request {
        if (action.type == "START_AUTHENTICATE" || action.type == "AUTHENTICATE"),
           let _ = action.payload
        {
            var headers = request.headers
            headers.removeValue(forKey: "Cookie")
            var urlParams = request.urlParams
            urlParams["ForceAuth"] = "true"
            let newRequest = Request(url: request.url, method: request.method, headers: headers, bodyParams: request.bodyParams, urlParams: urlParams, requestType: request.requestType, responseType: request.responseType, timeoutInterval: request.timeoutInterval)
            return newRequest
        }
        else {
            return request
        }
    }
}
```
This can be configured further to only run on specific Journeys. More details on request interceptors can be found [here](https://docs.pingidentity.com/sdks/latest/sdks/customize/how-to-customize-requests.html).

george-bafaloukas-forgerock:
Hello, a quick update that this problem has been addressed as part of the BETA release. [4.9.1-beta1](https://github.com/ForgeRock/forgerock-ios-sdk/releases/tag/4.9.1-beta1)

PR #376, opened 2026-04-30
fix: add cookie back after set sso token

harshvardhanarora:
# Description

The work done in [SDKS-3045](https://github.com/ForgeRock/forgerock-ios-sdk/pull/371) fixed a part of the issue where step up requests do not revoke the new session token but introduces a minor issue where the cookies received from the `/authorize` endpoint are first stored, and then deleted as a part of the `revokeSSOToken` in `SessionManager`. This causes all the subsequent requests made to be missing the session token cookie from the cookie store.

This change is not to be be considered for merging, only acts as a reference point to help developers point in the direction of the issue. The minor change addresses setting the cookie from the response post session token handling.

### DIFF
--- a/FRAuth/FRAuth/Authentication/AuthService.swift
+++ b/FRAuth/FRAuth/Authentication/AuthService.swift
@@ -215,7 +215,7 @@ public class AuthService: NSObject {
         // Invoke request
         FRRestClient.invoke(request: request, action: action) { (result) in
             switch result {
-            case .success(let response, _):
+            case .success(let response, let httpResponse):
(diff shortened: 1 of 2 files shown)


george-bafaloukas-forgerock:
Thanks for the report, this is solved as part of https://github.com/ForgeRock/forgerock-ios-sdk/pull/377 latest updates and should be available on the BETA release [4.9.1-beta1](https://github.com/ForgeRock/forgerock-ios-sdk/releases/tag/4.9.1-beta1)

PR #381, opened 2026-06-16
SDKS-4627 Harden device identifier persistence and add keychain diagnostics

george-bafaloukas-forgerock:
## Summary

Addresses **SDKS-4627 / TRIAGE-30702** — affected iOS users intermittently receive **different `forgerock.device.profile` identifiers for the same physical device**, causing the Device Profile Collector to fail device matching/rebinding. Root-cause analysis of the customer's SDK logs showed the identifier silently degrading to a **fresh random UUID on every call**, because the keychain key generation/storage and decryption were failing without surfacing any cause.

This PR hardens the identifier persistence path so a degraded device still yields a **stable** identifier, fixes a read-path bug that prevented self-healing, and adds detailed diagnostics so we can pinpoint the failure on devices we cannot reproduce locally.

## Background — how the bug manifests

`FRDeviceCollector` derives the device profile identifier from `FRDeviceIdentifier.getIdentifier()`, which resolves through four branches:

1. read the persisted identifier from the keychain (stable, expected)
2. re-derive it from the persisted public key (stable)
3. generate a new keypair, store it, derive the identifier (stable)
4. **else: hash a brand-new `UUID()` → different on every call**

On affected devices, branches 1–3 all fail and every call lands in branch 4, so each journey/collector produces a new identifier. The customer's 4.8.2 logs confirmed the cascade: `generateKeyPair()` → `"Failed to store Key Pairs into Keychain Service"` → UUID fallback. The earlier fix (PR #356, shipped in 4.8.4/4.8.5) only hardened the *write* side, so the customer remained affected.

### 1. Fix silent decrypt-failure in `KeychainService.getData` (the read-path bug)
When a `SecuredKey` is configured but `decrypt()` fails, the method previously returned the **raw, still-encrypted blob**. Callers then UTF-8-decoded garbage → `nil` → "identifier not found" → regenerate, *every time*. It now returns `nil` (treats the value as not-found) on decrypt failure, allowing the identifier to be regenerated once and re-stabilize. This is the path PR #356 never touched and the most likely reason 4.8.4/4.8.5 did not resolve the customer.

### 2. Retry transient keychain/Secure Enclave operations once
A `withRetry` helper wraps the keychain reads and keypair generation in `FRDeviceIdentifier`. A transient Secure Enclave failure (e.g. contention while a biometric prompt or other crypto op is in flight) no longer causes the SDK to skip a perfectly good persisted identifier. Deterministic failures still fail on both attempts, so the retry adds resilience without masking real errors.

### 3. Make the UUID fallback persist reliably
The fallback now persists **with retry and read-back verification** before returning, so even a degraded device returns the same identifier on subsequent calls instead of a new UUID each time.

### 4. Structured diagnostic logging across the full call chain
Adds `FRLog`/`Log` instrumentation (verbose/warning/error, respecting custom loggers) to:
- **`FRDeviceCollector`** — the identifier value placed into the profile
- **`FRDeviceIdentifier`** — which of the four resolution branches was taken, and whether persistence was verified
- **`KeychainService.set`/`getData`** — exact `OSStatus` codes with human-readable descriptions, encryption state, and a distinct warning when a value is found but fails to decrypt
- **`SecuredKey.init`/`readKey`/`decrypt`** — a loud warning when **no existing key is found and a new keypair is generated** (the signature of previously-encrypted data becoming permanently unreadable), plus decrypt failures on both primary and legacy algorithms
- **`KeychainManager`** — whether the `SecuredKey` initialized

On the next affected-device capture, the logs will disambiguate the competing hypotheses (Secure Enclave key loss vs. access-group/entitlement drift vs. device lock-state) from a single trace.

## Tests
- `test_getData_withUndecryptableValue_returnsNil` — writing with one SecuredKey and reading with another returns `nil`, not raw bytes
- `testWithRetry*` (×3) — first-attempt success, retry-then-succeed, and stop-after-one-retry
- `testDeviceIdentifierUUIDFallbackPersistsAndIsStable` — the fallback value persists and stays stable across calls

**All tests pass:** 30/30 on iPhone 17 (iOS Simulator 26.5), 0 skipped (`FRDeviceIdentifierTests`, `KeychainServiceTests`, `SecuredKeyTests`).

## Notes / caveat
The issue is not reproducible on our side; it is tied to specific physical devices. The retry only helps *transient* failures — if the affected devices fail because their Secure Enclave `SecuredKey` was regenerated (a *deterministic* decrypt failure), **change #1 is the fix** that breaks the regeneration loop. The new logging is intended to confirm the exact failure mode on the customer's devices.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/FRAuth/FRAuth/Device/FRDeviceCollector.swift
+++ b/FRAuth/FRAuth/Device/FRDeviceCollector.swift
@@ -2,7 +2,7 @@
 //  FRDeviceCollector.swift
 //  FRAuth
 //
-//  Copyright (c) 2019 - 2025 Ping Identity Corporation. All rights reserved.
+//  Copyright (c) 2019 - 2026 Ping Identity Corporation. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
@@ -49,7 +49,11 @@ public class FRDeviceCollector: NSObject {
         var result: [String: Any] = [:]
         result["version"] = FRDeviceCollector.FRDeviceCollectorVersion
         if let device = FRDevice.currentDevice {
-            result["identifier"] = device.identifier.getIdentifier()
+            let identifier = device.identifier.getIdentifier()
+            FRLog.v("FRDeviceCollector - collected device identifier for profile: \(identifier)")
+            result["identifier"] = identifier
+        } else {
+            FRLog.w("FRDeviceCollector - FRDevice.currentDevice is nil; device profile will not include an identifier")
         }
         for collector in self.collectors {
             dispatchGroup.enter()
--- a/FRAuth/FRAuth/Device/FRDeviceIdentifier.swift
+++ b/FRAuth/FRAuth/Device/FRDeviceIdentifier.swift
@@ -2,7 +2,7 @@
 //  FRDeviceIdentifier.swift
 //  FRAuth
 //
-//  Copyright (c) 2019 - 2025 Ping Identity Corporation. All rights reserved.
+//  Copyright (c) 2019 - 2026 Ping Identity Corporation. All rights reserved.
 //
 //  This software may be modified and distributed under the terms
 //  of the MIT license. See the LICENSE file for details.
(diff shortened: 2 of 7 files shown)


FRCore/FRCore/Keychain/KeychainService.swift
```diff
@@ -280,16 +306,35 @@ public struct KeychainService {
...
+                // If decryption fails, the data is unreadable with the current SecuredKey (e.g. the
+                // Secure Enclave key was regenerated). Returning the raw encrypted blob in that case
+                // would surface unusable bytes to callers (decoding to garbage / nil), so we treat it
+                // as not found and return nil instead.
+                if let decryptedData = securedKey.decrypt(data: returnedData) {
+                    Log.v("getData - retrieved and decrypted key '\(key)' (\(decryptedData.count) bytes)")
+                    return decryptedData
+                }
+                Log.w("getData - found \(returnedData.count) bytes for key '\(key)' but failed to decrypt with the current SecuredKey (service: '\(self.options.service)'); treating as not found")
+                return nil
```
vahancouver, line 328: in setData() we try to encrypt, and if we fail, we still save the unencrypted Data. However in getData() when we try to decrypt and fail, we just return nil. Seems like thos two are not aligned.

FRCore/FRCore/Keychain/KeychainService.swift
vahancouver, line 619: simplifyItems() still has the old broken pattern

  simplifyItems() still uses the original if let securedKey, let decryptedData = securedKey.decrypt(...) { returnedData = decryptedData } pattern — when decryption fails, the raw encrypted blob passes through to the caller. This is the exact bug that getData() was fixed to avoid. allItems() and debugDescription will still return garbage for undecryptable items.

FRAuth/FRAuth/Device/FRDeviceIdentifier.swift
```diff
@@ -52,38 +52,97 @@ public struct FRDeviceIdentifier {
...
+    ///
+    /// - Parameter operation: The operation to execute; returning nil indicates failure
+    /// - Returns: The first non-nil result, or nil if both attempts fail
+    static func withRetry<T>(_ operation: () -> T?) -> T? {
+        if let result = operation() {
+            return result
+        }
+        FRLog.v("Keychain operation failed; retrying once")
+        return operation()
+    }
```
vahancouver, line 159: The documented motivation is Secure Enclave contention. Two back-to-back attempts with no backoff provide little relief for contention. Even a 50ms fixed delay would make the retry semantically meaningful.

vahancouver (commented):
Please see the inline comments. There are two more comments from claude that I'm adding here (didn't add inline). See if they make sense to you.

1. testDeviceIdentifierUUIDFallbackPersistsAndIsStable does not test UUID fallback
  (FRAuth/FRAuthTests/.../FRDeviceIdentifierTests.swift)

  Deleting the three keychain keys and calling getIdentifier() on a properly initialized keychain service (with a working securedKey) exercises Branch 3 (successful key pair generation), not Branch 4 (UUID fallback). The specific regression this PR fixes — Branch 4 generating a new UUID on every call — has zero test coverage. The existing testDeviceIdentifierFallbackToUUID should be updated to assert that two sequential calls return the same value.

2. Branch 3 → Branch 4 race still possible (FRDeviceIdentifier.swift)

  If generateKeyPair() succeeds but both getData() retries in Branch 3 return nil, execution falls to Branch 4 (UUID returned). On the next call, Branch 2 succeeds (key pair is stored) and returns a key-derived identifier — different from the UUID returned on the first call. This residual case should be documented or Branch 3 should delete stored keys before falling through.

george-bafaloukas-forgerock:
Thanks for the thorough review @vahancouver — all four points are valid. Addressed in the follow-up commit ([b2b35a1](https://github.com/ForgeRock/forgerock-ios-sdk/commit/b2b35a1)):

**1. Encrypt/decrypt asymmetry in `set()` (inline comment):**
Changed the silent plaintext fallback to a hard `return false`. When a `SecuredKey` is configured, encryption failure now aborts the write entirely — storing plaintext would leave `getData()` unable to decrypt it and return `nil`, making the caller think the write succeeded but then fail to read it back. Consistent behaviour: if you can't write it safely, don't write it.

**2. `simplifyItems()` still had the old broken pattern (inline comment):**
Applied the same `nil`-on-decrypt-failure fix that `getData()` got. Undecryptable items are now skipped (with a warning log) so `allItems()` and `debugDescription` don't return raw encrypted blobs.

**3. No backoff on retry:**
Added a 50ms `Thread.sleep` before the retry attempt. The comment in `withRetry` explains the reasoning: two back-to-back calls provide no meaningful relief for Secure Enclave contention; a short fixed delay gives the coprocessor time to free up.

**4. `testDeviceIdentifierUUIDFallbackPersistsAndIsStable` didn't test Branch 4 / Branch 3→4 race:**
- Split into two tests. `testDeviceIdentifierUUIDFallbackPersistsAndIsStable` now forces Branch 4 via an invalid access group so all writes fail, and asserts the result is a well-formed 40-char SHA1 hex string. `testDeviceIdentifierUUIDFallbackIsStableWhenPersistenceWorks` pre-stores a UUID-derived identifier and verifies two consecutive calls return the same value via Branch 1.
- For the race itself: before entering Branch 4, `getIdentifier()` now calls `deleteExistingKeys()` and deletes the stored key-data blobs. This ensures any keys written by a successful `generateKeyPair()` that wasn't followed by a readable `getData()` are cleaned up, so a subsequent call can't Branch 2 into a key-derived identifier different from the UUID already returned.

spetrov (approved):
LGTM!

FRAuth/FRAuth/Device/FRDeviceCollector.swift
```diff
@@ -49,7 +49,11 @@ public class FRDeviceCollector: NSObject {
         var result: [String: Any] = [:]
         result["version"] = FRDeviceCollector.FRDeviceCollectorVersion
         if let device = FRDevice.currentDevice {
-            result["identifier"] = device.identifier.getIdentifier()
+            let identifier = device.identifier.getIdentifier()
+            FRLog.v("FRDeviceCollector - collected device identifier for profile: \(identifier)")
```
vahancouver, line 53: is it safe to log the identifier? should we truncate it or something?

FRAuth/FRAuth/Device/FRDeviceIdentifier.swift
```diff
@@ -52,38 +52,111 @@ public struct FRDeviceIdentifier {
...
+    /// errors.
+    ///
+    /// - Parameter operation: The operation to execute; returning nil indicates failure
+    /// - Returns: The first non-nil result, or nil if both attempts fail
+    static func withRetry<T>(_ operation: () -> T?) -> T? {
+        if let result = operation() {
+            return result
+        }
+        FRLog.v("Keychain operation failed; waiting 50ms before retrying once")
+        Thread.sleep(forTimeInterval: 0.05)
```
vahancouver, line 157: Do you think this sleep might become a problem? in the worst case it's called 4 times... it's 200 ms of potentially blocking a main thread. just something to consider.

vahancouver (approved):
Approved with 2 questions

FRAuth/FRAuth/Device/FRDeviceIdentifier.swift
george-bafaloukas-forgerock: should be OK, not ideal but no other way to do stop and wait as a last chance scenario

FRAuth/FRAuth/Device/FRDeviceCollector.swift
george-bafaloukas-forgerock: It's on `verbose` only and not PII right?

